PR #3, opened 2023-04-26
Update obj.def ITEM WEIGHTS

oldm8clint:
up to obj 805, line 8883


oldm8clint (commented):
will continue soon testing this all out :P

PR #21, opened 2023-07-10
Emotes

ultraviolet-jordan:
Animation


data/src/scripts/controls/player_controls.rs2
```diff
@@ -3,3 +3,46 @@
...
+
+[if_button,player_controls:com_24]
+~do_emote(emote_clap);
+
+[proc,do_emote](seq $anim)
+anim($anim, 0);
+// TODO: "You're already doing an emote!"
+def_int $client_frames = calc(seqlength($anim) * 20);
+def_int $delay = calc($client_frames / 600);
+p_delay($delay);
```
Jameskmonger, line 48: not too sure if it matters for `.rs2` but this file doesn't end in a newline

Jameskmonger (approved):
LGTM overall, one comment. Nice work

data/src/scripts/controls/player_controls.rs2
Pazaz: doesn't break the compiler, but would be good for consistency

ultraviolet-jordan: Fixed.

PR #23, opened 2023-07-11
refactor: conver NpcType and ObjType to ts

Jameskmonger:
Got a bit done on this, ObjType causes some errors in the `Player` class (`invSet`, `decodeIn`, `onCheat` and `invAdd`) functions

- Line `558`: `Argument of type 'ObjType' is not assignable to parameter of type 'ScriptArgument[]'`
- Line `662`: `Argument of type 'ObjType' is not assignable to parameter of type 'ScriptArgument[]'`
- Line `798`: `'objType' is possibly 'null'`
- Line `1946`: `Type 'number | undefined' is not assignable to type 'string | number'`

I'm pretty sure I've converted the TS correctly but I don't want to change engine code without getting some input 💬 

There are two causes:
- `ScriptRunner.init` is being passed an `ObjType` as a `ScriptArgument`. Is this permitted? It seems like `ScriptArguments` can only be `number` or `string`.
- `invAdd` and `invSet` call `ObjType.getId` and check for `-1` as an "invalid" state. `getId` can return `undefined`, we need to apply `?? -1` to the `getId` lookup if we want reliable failure results.    (probably need to do a bit of a healthcheck of the existing lookup functions, though TS conversions will help highlight these)

Will pick up any changes needed tomorrow, please let me know your suggested solutions and I can make them. Alternatively feel free to make the changes in JS and I can merge them into my branch for TS conversion.

### DIFF
--- a/src/lostcity/cache/NpcType.ts
+++ b/src/lostcity/cache/NpcType.ts
@@ -1,7 +1,8 @@
 import fs from 'fs';
 import Packet from '#jagex2/io/Packet.js';
+import { ConfigType } from './ConfigType';
 
-export default class NpcType {
+export default class NpcType extends ConfigType {
     static configNames = new Map();
     static configs: NpcType[] = [];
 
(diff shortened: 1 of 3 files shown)


Joshua-F:
The two `ScriptRunner.init` taking `objType` can just be `ScriptRunner.init(script, this)`. These scripts shouldn't be receiving any arguments.

Jameskmonger:
@Joshua-F I converted the two classes to use `ConfigType` base.

One issue: in `ObjType`, the decoding of `countco` relies on `count`, which isn't made available by the base `ConfigType`.

I could pass it in like this, thoughts?

```
        let dat = Packet.load(`${dir}/obj.dat`);
        let count = dat.g2();

        for (let id = 0; id < count; id++) {
            let config = new ObjType(id);
            config.decodeType(dat, count);
```

Jameskmonger:
Changes made, ready for review

Jameskmonger:
tested and working ✔️  ready for merge

image

PR #26, opened 2023-07-12
feat: pointer checking + small extras

Joshua-F:


src/lostcity/cache/FontType.ts
```diff
@@ -131,4 +131,42 @@ export default class FontType {
...
 		return size;
     }
+
+    split(str: string, maxWidth: number): string[] {
+        if (str.length === 0) {
+            // special case for empty string
+            return [str];
+        }
+
+        let lines: string[] = [];
```
Jameskmonger, line 141:
```suggestion
        const lines: string[] = [];
```

Doesn't appear to be reassigned (can do `.push` to a `const` no problem)

src/lostcity/engine/script/ScriptPointer.ts
```diff
@@ -0,0 +1,38 @@
+import { CommandHandler } from "#lostcity/engine/script/ScriptRunner.js";
+import ScriptState from "#lostcity/engine/script/ScriptState.js";
+
+/**
+ * Enumeration of possible pointer states used for runtime safety checks.
+ */
+enum ScriptPointer {
```
Jameskmonger, line 7: I find it's a good idea to add ordinal values to enums so that the values won't be changed automatically by adding new entries.

Here are some screenshots to show what I mean:

image
image

Jameskmonger (approved):
LGTM, couple of code style comments

src/lostcity/engine/script/ScriptPointer.ts
Joshua-F: In this case it isn't really relevant since the values won't be serialized anywhere and only matter at runtime.

Jameskmonger: Cool, yeah doesn't matter if it's never saved

Jameskmonger (approved):
LGTM ✅

PR #31, opened 2023-07-15
feat: implement a bunch of missing commands

Joshua-F:


src/lostcity/cache/LocType.ts
```diff
@@ -0,0 +1,198 @@
...
+
+        if (!fs.existsSync(`${dir}/loc.dat`)) {
+            console.log('Warning: No loc.dat found.');
+            return;
+        }
+
+        let dat = Packet.load(`${dir}/loc.dat`);
+        let count = dat.g2();
+
+        let active = -1;
```
Jameskmonger, line 22: This can be removed (you are now using `this.active` in the decoder)

src/lostcity/cache/LocType.ts
```diff
@@ -0,0 +1,198 @@
...
+        for (let id = 0; id < count; id++) {
+            let config = new LocType(id);
+            config.decodeType(dat);
+
+            if (active === -1) {
```
Jameskmonger, line 27: This should change to `config.active`

Jameskmonger: In fact I see you reassign `config.active` below for the Boolean value. You possibly need to split this into two values (but either way you arent setting the local `active`, from what I can see)

I looked into converting this file a couple of days ago and this active number/Boolean thing tripped me up too

Pazaz: Note that the active int vs Boolean is an important distinction in LocType:
https://github.com/2004scape/Client/blob/clean/client/src/main/java/jagex2/config/LocType.java#L251

It allows the config to override with a specific active=yes or active=no, or inherit default behaviors if not set (checks for -1). Active is used to determine if it can be interacted with/examined.

Pazaz: I switched it to use int on the type so I could merge this branch, thanks for the input. Retains original behavior as long as we don't check specifically `type.active === true`

PR #33, opened 2023-07-16
refactor: convert Inventory to typescript and store player invs in map

Joshua-F:


src/lostcity/engine/Inventory.ts
```diff
@@ -71,14 +77,14 @@ export class Inventory {
     stackType = Inventory.NORMAL_STACK;
 
     capacity = 0;
-    items = [];
+    items: (Item | null)[] = [];
```
Jameskmonger, line 80: May want to consider making this a `Map` in future too, I guess it depends on how much we will be accessing slot-based vs all items

Pazaz: Discussed this on our meeting last night. Will be reworking inventory more - have to implement a better listener system too to support player trading. Limited this issue's scope to be just conversion + small fixes

PR #37, opened 2023-07-16
refactor: corrects some item labels and separates talismans

HowSpooky:
wasn't too sure what to name the burnt fish, as multiple different fish use one given burnt id


Jameskmonger:
Could you name it based on the tool used to catch? I.e. burnt_fish_net, burnt_fish_fly_rod, etc?

Or are there still crossovers there

PR #74, opened 2023-07-31
docs: remove recommendation for NodeJS 19

Jameskmonger:
NodeJS 19 is already at end-of-life and has been for the last 2 months

NodeJS 18 is supported for another 2 years

We should not recommend odd-numbered users to anyone, ever. Only people specifically looking to be part of NodeJS beta channels.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ It won't ever profit off your trademarks. Without this project, this version of 
 
 ## Environment Dependencies
 
-- [NodeJS 18/19](https://nodejs.org/en)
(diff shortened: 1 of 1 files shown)


Pazaz:
thanks, 19 was mentioned for the people who had installed the latest ver. a year ago and didn't update since, so they know they don't have to upgrade. I'm not sure how anyone could install 19 now but it's good to clarify, so I'll merge

Jameskmonger:
It's still very possible to install Node 19 from their website or with NVM

PR #75, opened 2023-07-31
refactor: allow `false` value for `MEMBERS_WORLD`

Jameskmonger:
It was quite confusing that you had to either go for `true` or `not set`

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -4,7 +4,7 @@ GAME_PORT=43594
 
 LOCAL_DEV=true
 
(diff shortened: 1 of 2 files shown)


Pazaz:
.env files are parsed as strings, so =false is 'false' and becomes truthy in that ternary condition.
`console.log('false' ? 1 : 0);`
Maybe just compare to process.env.MEMBERS_WORLD === 'true'?

Jameskmonger:
changes pushed

PR #86, opened 2023-08-01
feat: updates to Restless Ghost

UnknownSeal:
* Reduced the amount of duplicated dialogue.
* Implemented untested spawning of the ghost when opening the coffin if not already present. Needs npc_add to be implemented to work.
* Ghost is removed when completing the quest.
* Made the Skull pickup go to normal pickup sub after intercepting.
* Changed the lost amulet check. Now only checks when the player is at the right quest step.


UnknownSeal:
Hmm I see that Hans apparently was added to the branch. That wasn't supposed to happen

Pazaz:
LGTM. Will merge anyways with Hans.

PR #97, opened 2023-08-02
Jest Test Packet class.

ultraviolet-jordan:


ultraviolet-jordan:
image

src/jagex2/io/Packet.test.ts
```diff
@@ -0,0 +1,239 @@
...
+    test('test jnstr', () => {
+        const string = 'Hello World!';
+        const expected = Packet.alloc(string.length + 1);
+        expected.pjnstr(string);
+        const result = new Packet(expected);
+        expect(result.gjnstr()).toBe(string);
+    });
+
+    test('test data', () => {
+        const random = randomBytes(255);
```
Jameskmonger, line 129: Generally I've found it's best to avoid using random (or other non-deterministic, e.g. datetimes) values in tests. It might be better to pre-compute some random data, so that this test executes the same way every time.

A lot of the painful tests I've dealt with in the past have been the ones using `random`/`time.now`

src/jagex2/io/Packet.test.ts
```diff
@@ -0,0 +1,239 @@
...
+        const result = new Packet(expected);
+        expect(result.g1s()).toBe(69);
+    });
+
+    test('test 1 signed b', () => {
+        const expected = Packet.alloc(1);
+        expected.p1(169);
+        const result = new Packet(expected);
+        expect(result.g1s()).toBe(-87);
+    });
```
Jameskmonger, line 40: Just a code style thing, so don't feel you need to address it here

But you can use `describe` to group similar tests together (even if you don't intend to share any behaviour)

```
describe("test 1", () => {
    test("unsigned", () => { /* ... */ });
    test("signed a", () => { /* ... */ });
    test("signed b", () => { /* ... */ });
});

describe("test 2", () => {
    test("unsigned", () => { /* ... */ });
    test("signed a", () => { /* ... */ });
    test("signed b", () => { /* ... */ });
});

// etc

Jameskmonger (approved):

Jameskmonger:
LGTM overall, nice

src/jagex2/io/Packet.test.ts
ultraviolet-jordan: Fixed. Thanks!

src/jagex2/io/Packet.test.ts
ultraviolet-jordan: Fixed. Thanks!

Jameskmonger (approved):

PR #98, opened 2023-08-02
build: add VS Code launch tasks for Jest debugging

Jameskmonger:
I had two issues with this:
- I can't use `integratedTerminal` (which I'd like to) but I think this might be some issue with my local setup (Unix-like shell in Windows environment)
- I can't reliably get console output from the tests

However besides that this seems to work well and uses the default VS Code inspector for stepping through the breakpoints.

Please test this branch on your machines and see if it all works for you too.


Jameskmonger:
image

Pazaz:
Works for me, nice! Going to merge now

PR #114, opened 2023-08-06
Fix midi_jingle command and add constants for jingles.

ultraviolet-jordan:


Pazaz:
death.mid and death2.mid were swapped in your PR - you might've thought this was a mistake, but according to some wiki members who received names from jagex, the actual player death jingle is `death2`. death is unused

Pazaz:
Oh actually, just noticed, you reorganized the jingles directory based on logical assumptions.

Let me go over the changes a little better:
1. firemarking and herblaw in unfortunately real.
2. "teasure hunt" was a typo so good catch
3. death2 is player death and they call it death2
4. they call the defence jingle, "defense"
5. for whatever reason they call it "duel win2" and not "duel win"

Pazaz:
other than that, if you revert the jingles directory changes, looks good to merge

PR #128, opened 2023-08-13
Cleanup collision code. Remove redundant code.

ultraviolet-jordan:
- Renamed 'blockproj' fields to 'blockrange'.
- Removed redundant code since TS can interpret enums from numbers.
- Adds `RoofCollider`.
- Adds `moverestrict=indoors` to Shop keepers and assistants.

image


src/lostcity/engine/collision/LocRotation.ts
```diff
@@ -1,18 +1,6 @@
...
-            case 2:
-                return LocRotations.EAST;
-            case 3:
-                return LocRotations.SOUTH;
-            default:
-                throw new Error('Invalid loc rotation.');
-        }
-    }
+export enum LocRotation {
+    WEST,
```
Jameskmonger, line 2:
```suggestion
    WEST = 0,
```

I would recommend to add constant values against each of them just to preserve them in case the enum ever gets reordered.

src/lostcity/engine/collision/LocShape.ts
```diff
@@ -1,88 +1,60 @@
...
-            case 21:
-                return LocShapes.ROOFEDGE_SQUARE_CORNER;
-            case 22:
-                return LocShapes.GROUND_DECOR;
-            default:
-                throw new Error('Invalid loc shape.');
-        }
-    }
+export enum LocShape {
+    WALL_STRAIGHT,
```
Jameskmonger, line 4: Likewise would recommend specifying constant values here

PR #152, opened 2023-09-18
Seers dialogues and other fixes

Indio3:
- dialogues for seer, bartender (seers village), and kebab seller, small inaccuracy fixes for banker/tower guard
- added grinding chocolate bars -> dust + fixed other messages
- fixed delay on hay searching


data/src/scripts/general_use/scripts/haybales.rs2
```diff
@@ -16,7 +16,7 @@ mes("You search the hay bail...");
 
 [label,search_haybale]()
 anim(bury, 0);
-p_delay(1);
+p_delay(3);
```
ultraviolet-jordan, line 19: This should be 3 ticks according to this:
image

`p_delay(2);`

data/src/scripts/general_use/scripts/spade.rs2
```diff
@@ -28,6 +28,4 @@ if(~trail_hasclue_inv = true) {
     }
 }
 
-~trail_clue_easy_reward;
```
ultraviolet-jordan, line 31: Is this supposed to be removed?

data/src/scripts/general_use/scripts/haybales.rs2
Indio3: didn't realize it was - 1, i'll change it

data/src/scripts/general_use/scripts/spade.rs2
Indio3: I think so, figured it was probably accidentally left in from testing (it'll send an easy clue reward on any dig where you don't have a map/coord clue in inv)

FlenarnTemp: Correct! It was used to generate up clue rewards during development, should be removed.

PR #173, opened 2023-10-07
Pirate's treasure

Desetude:
Dialogue primarily extracted from OSRS. I then compared with 2007 videos on the quest e.g. https://www.youtube.com/watch?v=SInMQzx1K1I.

Three notes:
1. I use the `scroll` interface for the pirate's message, which looks a bit sketchy. `trail_clue` is closer, but doesn't seem as appropriate.
image
image
2. I've added 4 varps to track different state in the quest. I could shrink this doing some bit magic, but seems a bit unnecessary?
3. I've split scripts between area & quest directories by whether the content would be used outside the quest (e.g. redbeard frank is only used for the quest so he's in `/quests/quest_hunt/`). Let me know if we've settled on a different approach


data/pack/varp.pack
```diff
@@ -301,3 +301,7 @@
 300=shop_sell
 301=shop_haggle
 302=thread_used
+303=plantation_employed
```
ultraviolet-jordan, line 304: Don't create new varps, use a varp id not listed from here please:
https://github.com/2004scape/Server/wiki/Varp-ID-Tracking

PR #197, opened 2023-10-23
feat: commands/headicons

ultraviolet-jordan:
- Adds the `headicons_set` command.
- Adds the `headicons_get` command.
- Adds the `headicon_add` proc.
- Adds the `headicon_del` proc.
- Adds the `headicon_exists` proc.
- Adds the `headicons_clear` proc.
- Fixes npc defend anim timing.


Pazaz:
should we flip the subject/action to: headicons_set headicons_get? most commands are like this format

PR #216, opened 2023-11-03
fix: script:symbols run script

ultraviolet-jordan:


Jameskmonger (approved):
I am a careless developer and @ultraviolet-jordan cleaned up my mess

LGTM

PR #227, opened 2023-11-28
fix: Secondary Pointers. HuntMode Packing. 

ultraviolet-jordan:
…xes.

### DIFF
--- a/src/lostcity/engine/script/handlers/NpcOps.ts
+++ b/src/lostcity/engine/script/handlers/NpcOps.ts
@@ -14,6 +14,7 @@ import NpcMode from '#lostcity/entity/NpcMode.js';
 import Player from '#lostcity/entity/Player.js';
 import Loc from '#lostcity/entity/Loc.js';
 import Obj from '#lostcity/entity/Obj.js';
+import HuntMode from '#lostcity/engine/hunt/HuntMode.js';
 
 const ActiveNpc = [ScriptPointer.ActiveNpc, ScriptPointer.ActiveNpc2];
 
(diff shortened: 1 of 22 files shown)


src/lostcity/tools/packconfig/NpcConfig.ts
```diff
@@ -208,8 +209,56 @@ export function parseNpcConfig(key: string, value: string): ConfigValue | null |
                 return null;
         }
     } else if (key === 'huntmode') {
-        // TODO
-        return value;
+        switch (value) {
```
Pazaz, line 212: This should be `PACKFILE.get('hunt')!.indexOf(value)` and return null if it's -1

Pazaz (changes requested):

src/lostcity/engine/script/handlers/NpcOps.ts
```diff
@@ -187,12 +188,22 @@ const NpcOps: CommandHandlers = {
     }),
 
     [ScriptOpcode.NPC_SETHUNTMODE]: checkedHandler(ActiveNpc, (state) => {
-        throw new Error('unimplemented');
+        const mode = state.popInt();
+        
+        if (mode === -1 || mode > HuntMode.BIGMONSTER_MELEE) {
```
Pazaz, line 193: Meant to include this in the review. Hunt lookups should go to the HuntType, and if the compiler needs to be aware of it then we just have to generate hunt.tsv values for the compiler (if it's not already)

Pazaz (approved):

PR #277, opened 2023-12-15
fix: Update mining text

boogie-nights:
Needed to make mining slightly passive aggressive about your abilities.

image
image


tannerdino:
Ah good catch. You might want to change the get_ore_fast too, iron uses that one.

tannerdino:
> Ah good catch. You might want to change the get_ore_fast too, iron uses that one.

Actually never mind, ill include some mining fixes in my next PR. Need to change mining to pass the $get_ore label instead of $get_ore proc

PR #280, opened 2023-12-18
feat: 🦠Plague City Quest & West Ardougne

ultraviolet-jordan:
- Missing cat selling to Civilians but it has the dialogue.
- Missing combat stats for some attackable mourners.


ultraviolet-jordan:
Real RuneScape images for reference material:

image
image
image
image
image
image
image
image
image
image
image
image

PR #306, opened 2023-12-27
feat: Added NPC hiscore migration

ipkpjersi:
Should the table name be plural or not? Who knows.

### DIFF
--- a/src/lostcity/db/types.ts
+++ b/src/lostcity/db/types.ts
@@ -24,8 +24,15 @@ export type newspost_category = {
     name: string;
     style: string;
 };
+export type npc_hiscore = {
+    id: Generated<number>;
+    npc_id: number;
+    account_id: number;
+    kill_count: Generated<number>;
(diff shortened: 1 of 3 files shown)


src/lostcity/db/types.ts
```diff
@@ -24,8 +24,15 @@ export type newspost_category = {
     name: string;
     style: string;
 };
+export type npc_hiscore = {
+    id: Generated<number>;
+    npc_id: number;
+    account_id: number;
+    kill_count: Generated<number>;
```
ipkpjersi, line 31: Is this supposed to be generated or should it just be number?

Pazaz: It's marked Generated because of the default value (so it doesn't always have to be specified on new inserts), good as-is!

ipkpjersi: Good to know, thanks.

PR #310, opened 2024-01-04
fix: Some more aggressive NPCs

boogie-nights:
This makes the world a lot deadlier, and harder to test without Single/Multi combat areas working properly. It could be a good idea to hold off on merging until that is completed.

* Updated some NPCs to use cowardly rather than aggressive_melee.
* * Cowardly stops the NPCs from attacking when the players are 2x their level, whereas aggressive_melee keeps them aggressive to the players at all times.
* * Did a quite run through of the F2P dungeon areas to check what needed to be aggressive or not. 
* Added animations to Barbarians


Pazaz:
Merging this with the understanding that deaths are coming... planning to touch on combat and issue reporting would be good.

PR #311, opened 2024-01-05
feat: Stealing Wine of Zamorak

boogie-nights:
Code for stealing Wine of Zamorak

Features:
* Reduces Attack, Strength, Defence, Ranged, Magic by 5%
* Damages player by their HP level / 20 + 1
* All monks in both zones attack player
* When all monks are dead the player can take the wine, but still receives damage (confirmed this happens on RS)
* Player can telegrab the wine

### DIFF
--- a/data/src/scripts/player/scripts/pickup.rs2
+++ b/data/src/scripts/player/scripts/pickup.rs2
@@ -4,7 +4,7 @@
 if (distance(coord, $coord) = 0) {
     @pickup_obj_floor($obj, $amount);
 } else {
-    @pickup_obj_table($obj, $amount);
+    @pickup_obj_table($coord, $obj, $amount);
 }
 
 [label,pickup_obj_floor](namedobj $obj, int $amount)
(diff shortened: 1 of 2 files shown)


data/src/scripts/player/scripts/pickup.rs2
```diff
@@ -15,10 +15,15 @@ anim(null, 0);
 obj_takeitem(inv);
 sound_synth(pick2, 0, 0);
 
-[label,pickup_obj_table](namedobj $obj, int $amount)
+[label,pickup_obj_table](coord $coord, namedobj $obj, int $amount)
 if (~pickup_obj_check_for_space($obj, $amount) = false) {
     return;
 }
+
```
Indio3, line 22: could do this in an opobj3 script instead:
```
[opobj3,wine_of_zamorak]
if (~wine_of_zamorak_attack(obj_type, obj_coord) = true) {
    return;
}
@pickup_obj(obj_coord, obj_type, obj_count);
```

boogie-nights: Now that's good thinking.

PR #328, opened 2024-01-28
feat: Shop Stock Updating

boogie-nights:
- Item Stock Increases if under set stock
- Item Stock decreases if over set stock
- Increases and decreases happen in accordance with stock rate property
- If the stock was not listed, e.g. the player sold to a general store then after 60 seconds the stock will decrease by one.
- Edited the InvConfig and InvType to write/read p4 to support the restock times needed.


src/lostcity/engine/World.ts
```diff
@@ -771,6 +771,34 @@ class World {
             }
 
             inv.update = false;
+
+            // Increase or Decrease shop stock
+            const invType = InvType.get(inv.type);
+            inv.items.forEach((item, index) => {
```
Pazaz, line 777: We should be checking `invType.restock == true` here so we're only running this logic on shops that (should) restock.
Removing items should probably be a 2nd loop if `invType.allstock == true` so we only run logic on shops that can stock player-supplied items.

src/lostcity/engine/World.ts
```diff
@@ -771,6 +771,34 @@ class World {
...
+                if (item) {
+                    // Item stock is under min
+                    if (item.count < invType.stockcount[index] && this.currentTick % invType.stockrate[index] === 0) {
```
Pazaz, line 780: More curiosity than anything since it looks fine -- if we do tick % stockrate here it'll always restore on a consistent boundary... does RS work like that? i.e. if something restocks every 5 mins and you buy at 4 min 55 s, does it restock in 5 s? or 5 min from the purchase time?

boogie-nights: Yeah, as strange as it seems that's how I observed it working on OSRS. For example, If you sold an item right before the 1 minute timer was up it would go on that refresh.

Pazaz: Makes sense then. Consider this comment resolved :) Just need to split up the logic above to use some more properties from InvType!

src/lostcity/engine/World.ts
boogie-nights: Ahh makes sense, I'll add the check around the restock. Do we want to go the two loop route? Wouldn't the shop that is allstock also be restock, so we'd be looping over the same thing twice since we're already in there for restock?

Pazaz: Hmm, I was leaving room for flexibility but it's probably a safe bet if `allstock=yes` then `restock=yes` as well. You can just wrap the last bit in `if (invType.allstock)` then

boogie-nights: Wrapped the code in restock, and wrapped the in allstock on the last bit of logic.

PR #329, opened 2024-01-29
fix: Player Controls and Auto Retaliate

boogie-nights:
In a recent PR Emotes, Auto-retaliate, and autocast selection were broken. These are some changes to preserve the intent (hopefully) and lets the content work again.

Additional Fixes:
- Updated XP for studded leather crafting
- Made the level check for studded leather crafting work


data/src/scripts/interface_controls/scripts/player_controls.rs2
```diff
@@ -29,7 +30,6 @@ queue(set_auto_retaliate, 0, 0);
 if_close;
 if (p_finduid(uid) = true) {
     %player_run = ^player_run_off;
-    return;
```
Pazaz, line 32: When you remove the return here the queue becomes redundant, %player_run is set to off then queues to set it to off again

boogie-nights: Ahh yep! That was an oversight left over from some testing. Pushed up the change for it just now.

Pazaz:
Some things have been fixed + main has diverged a bit so I'm going to take your one commit for studded and push it separately, thank you

PR #358, opened 2024-02-27
feat: Witches house quest

Indio3:
- witches house quest
- open_and_close_door2 proc for walk-through doors that instantly play the sound (seems to be quite a few of them, didn't change any existing in this pr)
- small fix for patrol mode, parsing huntrate

todo:
- a few things in the quest need to be changed to use huntall -> huntnext, couple of workarounds for now
- the op1 scripts for the mouse npc and mouse hole loc are not defined as I'm not sure what they do, and would prefer to not completely guess


src/lostcity/cache/HuntType.ts
```diff
@@ -91,7 +91,7 @@ export default class HuntType extends ConfigType {
         } else if (opcode === 10) {
             this.checkAfk = true;
         } else if (opcode === 11) {
-            this.rate = packet.g1();
+            this.rate = packet.g2();
```
ultraviolet-jordan, line 94: nice catch G

PR #379, opened 2024-04-02
fix: disable p2p equipment bonuses on f2p worlds

toomykins:


Jameskmonger (approved):

Joshua-F:
This shouldn't be necessary, but due to params loading _after_ objs atm the `autodisable` flag on the params are not being seen properly. I believe the proper fix would just be to load params before objs on startup.

PR #426, opened 2024-04-26
fix: prevent an additional object of the same id from being deleted when picking it up

### LINKED ISSUE #165: Objs of the same type on the same tile are treated as the same obj
Steps:
- Spawn two objs of the same id
- Drop them one at a time
- Pickup one of them
- You gain one and the other disappears.

o-k-a-y:
Fixes #165


Pazaz:
This code is eventually being replaced - maybe sooner rather than later after jkm’s next work gets packaged up and submitted - but I’m going to accept this change so this long-standing issue is fixed in the original implementation. Nice job!

PR #437, opened 2024-04-29
chore: husky OS compatibility

o-k-a-y:


Pazaz:
This fails on Windows, the syntax is invalid and uname isn't a command

o-k-a-y:
This should now work on Windows as well as Windows with WSL installed. There are two separate issues when using Github Desktop and this should also address issues with GitKraken and other GUI tools on Windows.

Pazaz:
```
.husky/pre-commit: line 5: return: can only `return' from a function or sourced script
husky - pre-commit script failed (code 2)
```
Testing on a fresh Windows install with pretty much only VS Code, husky spits out this error.

`exit` should work here, just replace both `replace`s

PR #456, opened 2024-05-02
fix: energy draining during exactmove

ultraviolet-jordan:


src/lostcity/entity/NetworkPlayer.ts
```diff
@@ -1110,7 +1110,7 @@ export class NetworkPlayer extends Player {
             }
         }
 
-        if (this.forceMove && pathfindX !== -1 && pathfindZ !== -1) {
+        if (this.delayed()) {
```
Jameskmonger, line 1113: Your changes  LGTM but as a future note we should probably move this into `Player` rather than `NetworkPlayer`.. I will make an issue

Jameskmonger (approved):

PR #463, opened 2024-05-04
refactor: netout property into a Stack

ultraviolet-jordan:


Pazaz:
We may want to use `Hashable` instead of `Linkable` - the original Packet source extended `Hashable` which extends `Linkable`

During Packet.release() -> .removeHead() affects the .next/.prev from Linkable due to it calling unlink. Linkable can only be part of 1 LinkList if I understand correctly? So if we want a LinkList on the player and a LinkList to pool allocations there's a good chance we won't be reusing memory as often.

Double check me - if I'm right then it needs to extend Hashable and we need a different container - either a 2nd LinkList that uses Hashable or maybe this Stack class? https://github.com/2004scape/Client2/blob/main/src/js/jagex2/datastruct/Stack.ts

Pazaz:
Hashable change was reverted, good. If we want this PR to get in then the container still needs to be changed from LinkList to something i.e. Stack for reasons above.

PR #514, opened 2024-05-19
fix(content): no longer remember attackstyles after logout

ultraviolet-jordan:
https://runescape.wiki/w/Update:Game_Improvements
`We've also rejigged the combat selection so now you'll stick to the same fighting style after you log out. This means that when you log in again, you'll automatically use your preferred style - no more defending yourself by hitting someone over the head with a staff when you have plenty of runes!`

### DIFF
--- a/data/src/scripts/_unpack/all.varp
+++ b/data/src/scripts/_unpack/all.varp
@@ -32,7 +32,6 @@ protect=no
 
 [attackstyle]
 transmit=yes
-scope=perm
 
 [com_stabattack]
 protect=no
@@ -128,7 +127,6 @@ transmit=yes
 
 [attackstyle_magic]
 transmit=yes
-scope=perm
 
 [varp_109]
 


tannerdino:
Revisiting this, this might be talking about autocast instead of attackstyles. Might need to find a video of someone logging in and opening up their combat menu before June 18 2007

vulprs:
I took some time to find uncut videos of login/welcome screen to combat tab. All examples I found showed the first attack style selected upon login.
https://youtu.be/WadjqZ_8OFs?t=134
https://youtu.be/ewSKBrpA8i0
https://youtu.be/Lej_vxEqYYg
https://youtu.be/1bRxgEJH4x0?t=62
https://youtu.be/mzS7CdZQEB0?t=46
https://youtu.be/HAeh4qE3WJ4?t=34

Although none of this is definitive proof, I think that last video is the most telling and it also lines up with my memory of: Log in -> Turn run on -> Switch attack style

PR #590, opened 2024-06-19
feat(content): Jungle Potion quest

lesleyrs:
~~WIP, sharing in advance so don't get duplicate work.~~

Done!


lesleyrs:
This is about as much as I can do, I was not able to find any old quest resources using web archive or google tools so it's most likely not completely accurate as it seems there were several changes in later RS2 (such as no seqs).

Left some TODO's for somebody to look at whoever still needs to do the quest on the official game. Also adding `huntmode=aggressive_melee` to tribesman NPC didn't seem to make it aggressive?

Besides that it's fully playable!

PR #628, opened 2024-07-06
refactor(engine): world

ultraviolet-jordan:
- Move debug commands from `ServerOps` to `DebugOps`
- Refactor `World` class and separate parts out of the cycle function.
- Create `WorldStat` and track last cycle and this cycle stats.
- World `invs` and `newPlayers` are refactored into a `Set`

### DIFF
--- a/data/src/scripts/_test/scripts/engine/debug_addobj.rs2
+++ b/data/src/scripts/_test/scripts/engine/debug_addobj.rs2
@@ -6,3 +6,25 @@ if (p_finduid(uid) = true) {
 } else {
     @please_finish;
 }
+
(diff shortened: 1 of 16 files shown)


data/src/scripts/_test/scripts/engine/debug_addobj.rs2
```diff
@@ -6,3 +6,25 @@ if (p_finduid(uid) = true) {
 } else {
     @please_finish;
 }
+
+[debugproc,addobj2](int $size)
```
Pazaz, line 10: Need local dev checks in commands that can add items

Pazaz (changes requested):

PR #670, opened 2024-07-17
feat(engine): save crc file for webworker + generate preloaddirs

lesleyrs:
This file is in the 225 runewiki cache so I assume it's original, and allows doing crc checks on webworker.

Simplified client2 changes as well: https://github.com/2004Scape/Client2/pull/21/files


Pazaz:
The crc file isn't "original" (included because people expect it) but it's original as in what Jagex would've served anyways yes :) that file is only ever in memory after requesting /crc.
Note these recent bundling changes reduce the flexibility of the engine and tie it more to a specific game version, some undoing/breaking may happen with the transition to 245 and beyond

lesleyrs:
readdir is now used to generate the arrays dynamically during bundling for the async functions.

PR #747, opened 2024-08-09
feat(content): duck quacking

tannerdino:


data/src/scripts/general/scripts/misc/sound.rs2
```diff
@@ -1,15 +1,11 @@
-[proc,sound_synth_area](synth $sound, int $delay, coord $coord, int $distance)
+[proc,sound_area](synth $sound, int $delay, coord $coord, int $distance)
 huntall($coord, $distance, 0);
 if (huntnext = true) {
```
Pazaz, line 3: Is this meant to be for 1 player or for all players around? `if (huntnext = true)` will run once, `while (huntnext = true)` will run to completion

tannerdino: Whoops, good catch, fixed

PR #748, opened 2024-08-09
fix(content): ~push_player proc for fm and mith seeds

tannerdino:
Also fixes an issue with teleports preventing movement for a tick


tannerdino:
Pretty certain this runescript should run 100% accurately. Obj interactions are weird in osrs... From the players turn (our `processInteractions()`), it seems like it repaths to the obj if the player hasnt moved?? Even if you're ontop of the obj

I assume theres some very specific engine behavior we're missing

PR #796, opened 2024-08-24
feat(content): slayer skill

hex-agon:


FlenarnTemp:
This looks great, but I'm not 100% sure on if this is something we would necessarily ever merge in this repository. As of right now, it'll most likely be another half year, if not more, before we even leave the 225 revision, and another good chunk of time before we're at January 2005 when Slayer is introduced.

This might be more suitable to close, and re-open when the time comes with focus being on content we have in the current revision, even more so as it's a bit unclear how the Github structure will be once we consider 225 complete.

PR #818, opened 2024-09-01
fix(content): junglepotion changes

Indio3:
corrected mesanims/bits of dialogue, freespace/stage check when collecting herbs, end stage is 12 instead of 11

### DIFF
--- a/data/src/scripts/skill_herblore/configs/herbs.obj
+++ b/data/src/scripts/skill_herblore/configs/herbs.obj
@@ -347,7 +347,7 @@ members=yes
 weight=7g
 param=identified_herb_level,3
 param=identified_herb_exp,25
-param=identified_herb_message,You identify the herb. It is Rogue's Purse.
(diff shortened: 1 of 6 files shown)


data/src/scripts/skill_herblore/configs/herbs.obj
```diff
@@ -347,7 +347,7 @@ members=yes
 weight=7g
 param=identified_herb_level,3
 param=identified_herb_exp,25
-param=identified_herb_message,You identify the herb. It is Rogue's Purse.
+param=identified_herb_message,You identify the herb. It is Rogues Purse.
```
ultraviolet-jordan, line 350: look at the video below this line change

Indio3: The obj name was changed from Rogues Purse to Rogue's Purse in August 2005, it seems like a safe assumption that this would of been updated then

PR #821, opened 2024-09-02
fix(content): Check cooking level when making chocolate cakes

vulprs:


data/src/scripts/skill_cooking/scripts/cooking_inv/scripts/cakes/cakes.rs2
```diff
@@ -101,6 +101,11 @@ if (inv_total(inv, $ingredient2) < 1) {
 return;
 
 [label,make_chocolate_cake](int $last)
+if (stat(cooking) < 50) {
+    // unsure on the authentic message, guessing
```
ultraviolet-jordan, line 105: why is it guessed

vulprs: I searched though a bunch of old videos trying to find an example of someone failing to make a chocolate cake, but didn't find one. I then looked at what we had for Pizzas and followed the same format. I've since checked OSRS and it matches what I've got here.
message
Should I change the comment to something like "OSRS message"?

Jameskmonger (approved):

PR #823, opened 2024-09-03
fix(content): big config update

thesuddensilent:
-some files sorted inside folder.
-falador.npc file configs is sorted by id.
-some wanderrange changed or added.
-some weapons have now new defend animations.
-2 seq animation name is renamed.
-legend guild shop is partially added.
-hores guild shop stacks and shop name changed matching new screenshot. -added some npc/object names on npc list.


data/src/scripts/quests/quest_zanaris/configs/quest_zanaris.obj
```diff
@@ -23,6 +23,7 @@ model=model_2648_obj
 2dxan=196
 iop2=Wield
 wearpos=righthand
+wearpos2=lefthand
```
ultraviolet-jordan, line 26: https://gitlab.com/waliedyassen/cache-dumps/-/blob/master/dump.obj?ref_type=heads#L9057

PR #827, opened 2024-09-04
fix(engine): unable to interact with bordering members locs on free world

ultraviolet-jordan:
- The fix is the `bordersFreeToPlay` function to check if a members loc borders a free zone, then we do not want to unload it.
- Did refactor `GameMap` and `World` to put `ZoneMap` inside of `GameMap`.

### DIFF
--- a/src/lostcity/engine/GameMap.ts
+++ b/src/lostcity/engine/GameMap.ts
@@ -15,10 +15,10 @@ import EntityLifeCycle from '#lostcity/entity/EntityLifeCycle.js';
 import Loc from '#lostcity/entity/Loc.js';
 import { Position } from '#lostcity/entity/Position.js';
 
-import Environment from '#lostcity/util/Environment.js';
-
 import { LocAngle, LocLayer } from '@2004scape/rsmod-pathfinder';
 import * as rsmod from '@2004scape/rsmod-pathfinder';
+import Zone from '#lostcity/engine/zone/Zone.js';
+import ZoneGrid from '#lostcity/engine/zone/ZoneGrid.js';
 
 export default class GameMap {
     private static readonly OPEN: number = 0x0;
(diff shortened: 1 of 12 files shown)


src/lostcity/engine/GameMap.ts
```diff
@@ -202,13 +235,13 @@ export default class GameMap {
...
                 const count: number = packet.g1();
-                if (!Environment.NODE_MEMBERS && !this.isFreeToPlay(absoluteX, absoluteZ)) {
+                if (!this.members && !this.isFreeToPlay(absoluteX, absoluteZ)) {
                     continue;
                 }
                 const objType: ObjType = ObjType.get(id);
                 const obj: Obj = new Obj(level, absoluteX, absoluteZ, EntityLifeCycle.RESPAWN, objType.id, count);
-                if (objType.members && Environment.NODE_MEMBERS || !objType.members) {
-                    zoneMap.zone(obj.x, obj.z, obj.level).addStaticObj(obj);
+                if (objType.members && this.members || !objType.members) {
```
Jameskmonger, line 243: just FYI you can get rid of the double-clause here with:

```suggestion
                if (!objType.members || this.members) {
```

Don't feel you have to change it for this PR though  just letting you know

src/lostcity/engine/GameMap.ts
```diff
@@ -243,10 +276,11 @@ export default class GameMap {
                 for (let z: number = 0; z < GameMap.Z; z++) {
                     const absoluteZ: number = z + mapsquareZ;
 
+                    if (!this.members && !this.isFreeToPlay(absoluteX, absoluteZ) && !this.bordersFreeToPlay(absoluteX, absoluteZ)) {
```
Jameskmonger, line 279: likewise this whole thing can be flipped if it makes the logic clearer

```suggestion
                    if (this.members || this.isFreeToPlay(absoluteX, absoluteZ) || this.bordersFreeToPlay(absoluteX, absoluteZ)) {
```

Jameskmonger (approved):
LGTM ✅

PR #869, opened 2024-09-15
fix(engine): engine tweaks to match osrs firemaking edgecases

tannerdino:
Engine:
- extra check to decrement waypoint if player is standing on waypoint tile in . Currently our waypoints only decrement on a change of direction during the players path. P_teleports can cause this change of direction calculation to change, as p_teleport can change the player's x and z before movement is processed. So an additional check is necessary
- clearWaypoints *before* executing the script in `processInteractions()`, unless aprange is called. If a p_op command is called within the aptrigger, clearWaypoints after executing the script (Not entirely sure if this is how it should be implemented, but thats how the old code handled p_op commands in aptriggers)
- clearWaypoints *before* executing the script for npc's in `aiMode()`
- still queueWaypoint to current tile if current tile and target tile are the same (todo: update rsmod's findpath function)
- adjust moveToClick & op handler behavior to not clearWaypoints when delayed
- don't process movement if busy & have something queued, instead of just having modal interface open and something queued

Content:
- south and north directions for ~push_player were flipped
- anim(null, 0) should be called before ~push_player instead of after
- debugprocs for testing interactions

[Firemaking edgecases that are now 1:1](https://github.com/2004Scape/Server/wiki/Firemaking-Edge-Cases)

# Some Interactions that remain unaffected
## Pathing to entities as they delete
https://github.com/user-attachments/assets/8f74a2b4-5fde-4801-8bed-c5e123495104

## Pathing to npc target as npc is changing type:
https://github.com/user-attachments/assets/da823480-6f3a-4185-9114-f2c53310d722

## ai_opplayer interactions while moving
https://github.com/user-attachments/assets/3d0b94c1-d10e-423a-b294-db8bb4dbaf7d

## ai_applayer interactions while moving
https://github.com/user-attachments/assets/cc5f9ca1-5dc8-4254-bd45-f0f9a40a0190

## Repathing on last waypoint
https://github.com/user-attachments/assets/0f120541-fc81-42b1-8591-5466a0ee7751

## Not repathing
https://github.com/user-attachments/assets/84767c30-7c4c-468e-93c6-af6761d794a9

### DIFF
--- a/src/lostcity/entity/Player.ts
+++ b/src/lostcity/entity/Player.ts
@@ -532,7 +532,7 @@ export default class Player extends PathingEntity {
 
     updateMovement(repathAllowed: boolean = true): boolean {
         // players cannot walk if they have a modal open *and* something in their queue, confirmed as far back as 2005
-        if (this.containsModalInterface() && this.queue.head() != null) {
+        if (this.busy() && this.queue.head() != null) {
             this.recoverEnergy(false);
             return false;
         }
--- a/src/lostcity/entity/Player.ts
+++ b/src/lostcity/entity/Player.ts
@@ -918,14 +918,19 @@ export default class Player extends PathingEntity {
         } else if (apTrigger && this.inApproachDistance(this.apRange, this.target)) {
             const target = this.target;
             this.target = null;
+            const wayPoints = this.waypoints;
+            const waypointIndex = this.waypointIndex;
+            this.clearWaypoints();
 
             this.executeScript(ScriptRunner.init(apTrigger, this, target), true);
 
             // if aprange was called then we did not interact.
             if (this.apRangeCalled) {
+                this.waypoints = wayPoints;
+                this.waypointIndex = waypointIndex;
                 this.target = target;
             } else {
-                if (this.target === target) {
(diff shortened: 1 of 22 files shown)


src/lostcity/entity/Player.ts
```diff
@@ -532,7 +532,7 @@ export default class Player extends PathingEntity {
 
     updateMovement(repathAllowed: boolean = true): boolean {
         // players cannot walk if they have a modal open *and* something in their queue, confirmed as far back as 2005
-        if (this.containsModalInterface() && this.queue.head() != null) {
+        if (this.busy() && this.queue.head() != null) {
```
Pazaz, line 535: `Player has interface open and queue is active`
->
`Player has interface open or is delayed, and queue is active` (busy)

Do you want to this temporarily disable movement if the player is delayed and another script queues before it finishes?
I’m just wondering if there’ll be any problems with p_walk / delay logic.

tannerdino: I think an additional check is needed for redx and yellowx stuff, to prevent p_walk + p_delay issues.

PR #878, opened 2024-09-17
fix(engine): save all players and shut down on World cycle error

Jameskmonger:


tannerdino:
Technically inauthentic? https://oldschool.runescape.wiki/w/Server_crash_glitch

Jameskmonger:
> Technically inauthentic? https://oldschool.runescape.wiki/w/Server_crash_glitch

don't think so - IMO that relates to hardware issues/OOMs etc (as the wiki page specifically mentions intensive actions involving hundreds of accounts) rather than logic errors in code

PR #879, opened 2024-09-17
fix(engine): remove `undefined` from `EntityList` and `ZoneEntityList`

Jameskmonger:
not sure of the purpose of these - but they lead to some invalid TypeScript (pretty sure it's the cause of CI failures)


Jameskmonger:
not needed for now, fixed in #880

we will have some changes to make to these iterators when we go to TS 5.6+

PR #906, opened 2024-09-25
fix(content): authentic messages

stomponbabyheads:
- Update cooking on fire message
fire
- Update woodcutting level-up messages
yew
- Update [Bowstring message](https://youtu.be/HQEpeWAUMbI?si=1OHnT82XJvqdt3TT&t=49)
- [Update message for trying to pick up someone's cat](https://youtu.be/WsvTr-1bZwM?t=175)
- Update coloring on low level message for Prayer

### DIFF
--- a/data/src/scripts/interface_chat/scripts/chat.rs2
+++ b/data/src/scripts/interface_chat/scripts/chat.rs2
@@ -550,6 +550,17 @@ while ($page < $pagetotal) {
     $page = calc($page + 1);
 }
 
+// imgur.com/WX3hzXI
+[proc,objboxsmall](obj $obj, string $string)
+split_init($string, 400, 4, q8);
+def_int $page = 0;
+def_int $pagetotal = split_pagecount;
+while ($page < $pagetotal) {
+    ~objbox_page($obj, 180, 0, divide(^objbox_height, 3), $page);
+    p_pausebutton;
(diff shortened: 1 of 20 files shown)


data/src/scripts/interface_chat/scripts/chat.rs2
```diff
@@ -550,6 +550,17 @@ while ($page < $pagetotal) {
     $page = calc($page + 1);
 }
 
+// imgur.com/WX3hzXI
+[proc,objboxsmall](obj $obj, string $string)
```
ultraviolet-jordan, line 554: There is no such thing as an `objboxsmall`. This must be reverted. The objboxes were individually customized during this time I've only explained it 50 times to people. Either use the already existing objboxes or delete the custom ones entirely.

ultraviolet-jordan: https://github.com/2004Scape/Server/wiki/FAQ#some-item-chat-boxes-are-off-center

stomponbabyheads: Chill, it's just a temporary solution. you are right and that is why my next PR will replace all instances of objboxt, objboxb, objboxwhathaveyou, and add an objbox with 3 more arguments.

PR #908, opened 2024-09-26
feat(content): Make ice gloves accessible

kylmp:
First time messing with your runescript, so let me know if anything can be done better!


kylmp:
Just noticed you should be able to use a pickaxe on the rockslide too, so I need to add that

data/src/scripts/skill_mining/scripts/pickaxe_checker.rs2
```diff
@@ -19,4 +19,45 @@ if ($lvl0 >= oc_param(iron_pickaxe, levelrequire) & ($obj1 = iron_pickaxe | inv_
 if ($lvl0 >= oc_param(bronze_pickaxe, levelrequire) & ($obj1 = bronze_pickaxe | inv_total(inv, bronze_pickaxe) > 0)) {
     return(bronze_pickaxe);
 }
-return(null);
\ No newline at end of file
+return(null);
+
+// return int: 1 = valid pickaxe, 2 = does not have level to use the pickaxe, 3 = not a pickaxe
```
Pazaz, line 24: In the future I would recommend utilizing constants (declare them inside a .constant file as `^name = value`) so you can do e.g. ^valid_pickaxe, ^unusable_pickaxe, ^invalid_pickaxe instead of a magic number, it helps when the codebase grows

data/src/scripts/areas/area_white_wolf_mountain/scripts/white_wolf_mountain.rs2
```diff
@@ -0,0 +1,40 @@
...
+p_delay(1);
+loc_change(loc_474, 2);
+p_delay(1);
+loc_del(3);
+loc_add($loc_coord, loc_475, loc_angle, loc_shape, 3); // replace rockslide with 2x2 invisible barrier
+loc_add(movecoord($loc_coord, 0, 0, 1), loc_475, loc_angle, loc_shape, 3);
+loc_add(movecoord($loc_coord, 1, 0, 1), loc_475, loc_angle, loc_shape, 3);
+loc_add(movecoord($loc_coord, 1, 0, 0), loc_475, loc_angle, loc_shape, 3);
+if (coordx(coord) > coordx($loc_coord)) {
+  ~forcemove(movecoord($loc_coord, -1, 0, 1));
```
Pazaz, line 37: I haven't confirmed, but OSRS takes a slightly different path: https://youtu.be/NVsE2iT4eEU?t=105 - walks two tiles forward and then diagonally down. Your video walks diagonally down then forward.
I wonder if p_walk or a similar proc would look different than forcemove? (internally it's a chain of teleports)

Just a note for whoever takes up Hero's Quest and takes a look at this in the future. Not a blocker.

Pazaz:
I combed over it and added a couple comments, looks good for a first time :) I would also test with another player logged in and watching that player, double check the invisible barrier works and things look correct on both sides. I'll pull your branch down later and check then if no one else approves it.

data/src/scripts/areas/area_white_wolf_mountain/scripts/white_wolf_mountain.rs2
Indio3: based off https://youtu.be/2M-W0tUicdw?si=7Q3NIVHW099pNJ2_&t=181 it should probably be forcemove(movecoord(coord, 2, 0, 0)) and p_walk(movecoord(coord, 1, 0, -1))

Indio3: https://github.com/user-attachments/assets/a070ab69-a04e-4d75-91d1-37887bba1920

data/src/scripts/areas/area_white_wolf_mountain/scripts/white_wolf_mountain.rs2
```diff
@@ -0,0 +1,40 @@
+[oploc1,ice_queen_lair_rockslide]
+mes("These rocks contain nothing interesting. They are just in the way.");
+
+[oploc2,ice_queen_lair_rockslide]
+def_obj $pickaxe = ~pickaxe_checker;
+if ($pickaxe = null) {
+  ~mesbox("You need a pickaxe to clear the rockslide. You do not have a pickaxe which you have the Mining level to use.");
```
Indio3, line 7: this is a slightly different mes on OSRS, or do you have an old source with this mesbox?: 

`message_game | type=gamemessage, message="You need a pickaxe to clear the rockslide. You do not have one that you have the Mining level to use."`



</body>
</html>

data/src/scripts/skill_mining/scripts/pickaxe_checker.rs2
kylmp: Yes will do that, I didn't like how I did it but couldn't think of another way

data/src/scripts/areas/area_white_wolf_mountain/scripts/white_wolf_mountain.rs2
```diff
@@ -0,0 +1,40 @@
...
+} else {
+  ~mine_ice_queen_lair_rockslide($pickaxe);
+}
+
+[oplocu,ice_queen_lair_rockslide]
```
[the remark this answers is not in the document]
kylmp: Ah much better, thanks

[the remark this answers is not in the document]
kylmp: Actually, this would result in say using a steel pick on the rock but if you have rune in inv it would use the rune one. Need to check how this works in game to see if it would use the one you used on the rock or your best in inv like normal.

data/src/scripts/areas/area_white_wolf_mountain/scripts/white_wolf_mountain.rs2
kylmp: I will change it to that, I assumed the second sentence would be the same as other pickaxe checkers

data/src/scripts/areas/area_white_wolf_mountain/scripts/white_wolf_mountain.rs2
kylmp: From your OSRS vid it seems like it is just player location -> 2 steps (west/east) -> 1 step (up/down)

kylmp: Made some changes to match OSRS which I guess is our best thing to go off of 

https://github.com/user-attachments/assets/a7602928-35c6-4586-918a-a45584abe68b

data/src/scripts/areas/area_white_wolf_mountain/scripts/white_wolf_mountain.rs2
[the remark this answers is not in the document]
tannerdino: The oploc interaction would be processed next tick from the p_oploc(2) as well, need to check if your character does nothing for a tick in osrs when using a pickaxe on it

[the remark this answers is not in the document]
kylmp: So in OSRS it does use your best pickaxe regardless of which you use on the rocks. It does not appear to add any extra delay though

kylmp:
Ok ready for re-review now!

data/src/scripts/areas/area_white_wolf_mountain/scripts/white_wolf_mountain.rs2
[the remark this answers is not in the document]
kylmp: Just noticed your comment > it has a 1t delay for the oplocu interaction

How do you find that? In any case i'll update it as you suggested then, though with the delay I think the mining animation might be too short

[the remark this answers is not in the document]
Indio3: I mean that there is a 1t delay between when you use the pickaxe on the rock, and when the action starts, which is because p_oploc(2) is being called, wwill run on the next tick, causing a 1t delay and matching the behaviour where the best pickaxe is used:

https://github.com/user-attachments/assets/9d221c68-2e4f-436f-9bba-ec401bf367b3

compare:
```
Tick 346 |  
-- | --
oploct | id=2634, coord=(2838, 3517, 0), com=149:0, sub=2, obj=19564
message_game | type=gamemessage, message="Nothing interesting happens."
```

vs 
```
Tick 349 |  
-- | --
oploct | id=2634, coord=(2838, 3517, 0), com=149:0, sub=1, obj=1275
player_info |  
idle | player=(index=211, name=Indio2, coord=(2837, 3517, 0))
face_angle | angle=1432
Tick 350 |  
varp_large | id=1042, old=28928, new=28928
player_info |  
idle | player=(index=211, name=Indio2, coord=(2837, 3517, 0))
sequence | id=7139
```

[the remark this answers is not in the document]
kylmp: oh right forgot to update, updated now

Indio3 (approved):

PR #909, opened 2024-09-27
fix(content): more authentic messages, level up unlocks

stomponbabyheads:
- Update some messages to match old screenshots.
  - Oracle
  - Witch
  - Fishing guild shopkeeper
  - Catching rats
- Remove black punctuation from some level up unlocks to match old screenshots
  - Defence
  - Mining
  - Fishing
    - Add missing manta ray and sea turtle messages
- Fix 2 typos in dwarf cannon manual


data/src/scripts/levelup/scripts/levelup_unlocks_fishing.rs2
```diff
@@ -1,21 +1,21 @@
...
+    case 25: ~doubleobjbox(raw_pike,fishing_rod,"You can now try fishing for @dbl@Pike.");
+    case 30: ~doubleobjbox(raw_salmon,fly_fishing_rod,"You can now try fishing for @dbl@Salmon.");
+    case 35: ~doubleobjbox(raw_tuna,harpoon,"You can now try harpooning @dbl@Tuna.");
+    case 40: ~doubleobjbox(raw_lobster,lobster_pot,"You can now try to catch @dbl@Lobsters.");
+    case 46: ~doubleobjbox(raw_bass,net_big,"Members can now try fishing for @dbl@Sea Bass with a large net.");
+    case 50: ~doubleobjbox(raw_swordfish,harpoon,"You can now try harpooning @dbl@Swordfish.");
+    case 53: ~doubleobjbox(raw_lava_eel,fishing_rod,"Members can now try catching @dbl@Lava Eels.");
+    case 68: ~mesbox("Members are now qualified to enter the @dbl@Fishing Guild.");
+    case 76: ~doubleobjbox(raw_shark,harpoon,"Members can now try harpooning @dbl@Sharks."); //imgur.com/SvV4b4p
+    case 79: ~doubleobjbox(raw_sea_turtle,bucket_empty,"Members can now catch @dbl@Sea Turtles on the Trawler."); //imgur.com/T0raUc4
```
ultraviolet-jordan, line 19: No offense dawg but these are black not blue. Only like the dragon one was blue.... Doesn't call for changing every single one of these as well................

stomponbabyheads: >Doesn't call for changing every single one of these as well................

It's the same logic as capitalising all of them, no................? It's a very safe assumption considering all the screenshots (2004-2024) we have................dawg................

ultraviolet-jordan: It's not the same logic. Every skill unlock is capitalized, you would be suggesting that every skill unlock *for all skills* would use a blue colored punctuation at the end, which wouldn't be right.

PR #910, opened 2024-09-29
feat(content): heros quest

kylmp:
The black arm portion definitely needs the most verification, my only reference is YT videos and my OSRS acc is phoenix gang. 

Started documenting references I found, but should have done it more. 

Some of the dialogue options that no one intentionally clicks on are hard to find / not totally complete... however, anything that I was not sure of I added a TODO comment nearby


data/src/scripts/quests/quest_hero/scripts/brimhaven_blackarm.rs2
```diff
@@ -0,0 +1,265 @@
...
+[oploc1,grips_cupboard_opened]
+@summon_grip(false);
+
+[label,summon_grip](boolean $open_cabinet)
+if ($open_cabinet = true) {
+  loc_change(grips_cupboard_opened, 100); // TODO - how long to keep open?
+}
+npc_findall(coord, pirate_guard, 3, 0);
+while(npc_findnext = true) {
+  if(npc_type = pirate_guard) { // TODO - this needs to be the closest guard to the player, currently selects furthest/random?
```
kylmp, line 218: other than dialogue, this was one technical part I didn't figure out. When you search this cupboard you need to face the closest pirate guard to you based off of videos.

This code here seems to pick the pirate guard furthest away (theres a couple of them walking around)

tannerdino: You can use the `npc_find` engine command

data/src/scripts/quests/quest_hero/configs/quest_hero.constant
```diff
@@ -0,0 +1,16 @@
+^hero_not_started = 0
```
Indio3, line 1: The quest stages here don't match up with OSRS (https://chisel.weirdgloop.org/varbs/display?varplayer=188, 15 is completion, 1-6 for phoenix and 7-13 for black arm) if you're completing the quest on OSRS you can use rsprox to dump info like this

data/src/scripts/quests/quest_hero/scripts/brimhaven_blackarm.rs2
kylmp: Ah thats... much better. Not sure why I used findall multiple times but replaced them all now, and that did fix the issue about choosing the furthest guard

data/src/scripts/quests/quest_hero/configs/quest_hero.constant
kylmp: Didn't realize this was known to such an extent. I'll see how different it is if it strays too much maybe should just scrap most of the blackarm/phoenix portions

kylmp: I suppose theres no way to know what each individual stage is without actually doing it on osrs with rsprox?

PR #921, opened 2024-10-08
fix(engine): last_coord command to use last changed coord instead of last step

tannerdino:
also refactors wildy warning code to only show wildy warning when you pass the "wilderness line"

### DIFF
--- a/src/lostcity/entity/PathingEntity.ts
+++ b/src/lostcity/entity/PathingEntity.ts
@@ -53,6 +53,9 @@ export default abstract class PathingEntity extends Entity {
     lastInt: number = -1; // resume_p_countdialog, ai_queue
     lastCrawl: boolean = false;
     lastMovement: number = 0;
+    lastCoordX: number = -1;
+    lastCoordZ: number = -1;
+    lastCoordLevel: number = -1;
 
     walktrigger: number = -1;
     walktriggerArg: number = 0; // used for npcs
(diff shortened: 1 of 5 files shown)


src/lostcity/entity/PathingEntity.ts
```diff
@@ -53,6 +53,9 @@ export default abstract class PathingEntity extends Entity {
     lastInt: number = -1; // resume_p_countdialog, ai_queue
     lastCrawl: boolean = false;
     lastMovement: number = 0;
+    lastCoordX: number = -1;
```
ultraviolet-jordan, line 56: What's the difference between this and lastTickX?

tannerdino: lastTickX is reset every tick whereas lastCoordX is reset only if the player moves/teleports

ultraviolet-jordan: I see. How come lastStepX isn't good for you?

tannerdino: If using lastStep: For the wilderness warning, checking if the last_coord was south of the wilderness warning zone can be skipped if you're running. Since when you're running you step twice in the same tick, last step wont be south of the wilderness warning zone.

Work around would be the previous implementation, where it checks if the current coord is north of the last_coord. This spams the wilderness warning if moving northwest/northeast between west and east zones. Big issue during black knights fortress since thats right on the wilderness warning zone line

tannerdino: Its probably worth asking ash if a last_coord command exists

ultraviolet-jordan: Yeah worth asking him.
Maybe this is an area queue or something like that is last_coord doesn't exist.

PR #925, opened 2024-10-10
feat(content): east ardy sewer gate

kylmp:
- make east ardy sewer gate openenable/picklockable
- fix issue with not be able to reobtain ID papers from trobert during heroes quest

https://github.com/user-attachments/assets/cdca8027-4a6f-4ca7-95d5-19ba4087565a

### DIFF
--- a/data/src/scripts/skill_thieving/scripts/doors/locked_door.rs2
+++ b/data/src/scripts/skill_thieving/scripts/doors/locked_door.rs2
@@ -111,22 +111,12 @@ if ($entering = true) {
         }
     }
 
-    mes("You attempt to pick the lock.");
-
     def_int $current_level = stat(thieving);
     def_int $thieving_level = db_getfield($data, locked_door:level, 0);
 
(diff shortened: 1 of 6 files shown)


data/src/scripts/skill_thieving/scripts/doors/locked_door.rs2
```diff
@@ -111,22 +111,12 @@ if ($entering = true) {
...
-        return;
-    }
-
-    def_int $roll = random(100);
-    if ($roll < 50) { // Some arbitrary rng.
-        mes("You fail to pick the lock.");
+    if (~attempt_picklock_door($current_level, $thieving_level) = false) {
         return;
     }
-    mes("You manage to pick the lock.");
```
kylmp, line 129: extracted to new proc for reuse

kylmp: In that case it probably does, seemed like a bug in the script with an unreachable condition

PR #931, opened 2024-10-11
fix(content): Ring of Wealth doesn't affect mega rare drop table

Indio3:
also added legends quest requirement to accessing mega rare table from gem drop table

### DIFF
--- a/data/src/scripts/drop tables/scripts/shared_droptables.rs2
+++ b/data/src/scripts/drop tables/scripts/shared_droptables.rs2
@@ -63,11 +63,11 @@ if ($random < 32) {
     if (map_members = true) {
         return (half_key2, 1);
     }
-} else if ($random < 62) {
+} else if ($random < 62 & %legends_progress = ^legends_complete) {
     if (map_members = true) {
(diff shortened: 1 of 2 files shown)


data/src/scripts/drop tables/scripts/shared_droptables.rs2
```diff
@@ -63,11 +63,11 @@ if ($random < 32) {
     if (map_members = true) {
         return (half_key2, 1);
     }
-} else if ($random < 62) {
+} else if ($random < 62 & %legends_progress = ^legends_complete) {
     if (map_members = true) {
-        // TODO - Only go to mega-rare if player has completed Legends' Quest, else give a Talisman.
+        return (~megararetable);
     }
-} else if ($random < 65) {
```
kylmp, line 70: Is that right? Without a wealth ring you'll get a talisman from here half the time?

tannerdino: Yeah does look like it removes the 63/128 chance of dropping nothing

Indio3: oh good point, will update

PR #952, opened 2024-10-22
fix: Revised non-standard crossbow bolt ranged strength bonus

nerium-oleander:
Per discussion in discord sources channel (username is teagan) revised non-standard crossbow bolts to give a +18/+9 each tier /half-tier, as there is some indirect evidence that this schema pre-dates the crossbow rework in 2006. Consolidated comments to one longer comment, which explains the change.


ultraviolet-jordan:
https://github.com/user-attachments/assets/3a9bb610-ceee-4baa-b4d4-a0618180a42e

PR #954, opened 2024-10-23
fix(content): inverted locked doors

kylmp:
A few doors have the locked side as the inverse of the entering side:
door 1 - pirate hut wildy (pos 3045, 3956)
door 2 - yanille dungeon (pos 2601, 9481)

### DIFF
--- a/data/src/scripts/skill_thieving/scripts/doors/locked_door.rs2
+++ b/data/src/scripts/skill_thieving/scripts/doors/locked_door.rs2
@@ -23,10 +23,10 @@
 ~attempt_open_locked_door(loc_type, ~check_axis_locactive(coord));
 
 [oploc1,loc_2558]
-~attempt_open_locked_door(loc_type, ~not_bool(~check_axis_locactive(coord)));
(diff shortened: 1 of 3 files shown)


data/src/scripts/skill_thieving/scripts/doors/locked_door.rs2
```diff
@@ -54,27 +54,27 @@
 
 [oplocu,loc_2557]
 if(last_useitem = lockpick) {
-    ~pick_locked_door(loc_type, loc_coord, ~not_bool(~check_axis_locactive(coord)));
+    ~pick_locked_door(loc_type, loc_coord, ~check_axis_locactive(coord));
```
kylmp, line 57: this one was just a misconfiguration, it is not an inversed door (axe hut in deep wildy)

ultraviolet-jordan:
I hope that this isn't related to something bigger like it depending on the rotation of the `Loc` or something. Maybe the code works for those locs already but at other parts of the map.

PR #955, opened 2024-10-23
fix(engine): yellow x runs walktrigger multiple times per tick

tannerdino:
matches:
- late 2006: https://youtu.be/Hsfn5HHtbSk?t=125
- early osrs: https://youtu.be/dKfFkEyqD8E?t=9

### DIFF
--- a/src/lostcity/engine/World.ts
+++ b/src/lostcity/engine/World.ts
@@ -691,11 +691,12 @@ class World {
                             player.pathToPathingTarget();
                             continue;
                         }
-        
-                        player.pathToMoveClick(player.userPath, !Environment.NODE_CLIENT_ROUTEFINDER);
-    
-                        if (!player.opcalled && player.hasWaypoints()) {
-                            player.processWalktrigger();
+                        if (!Environment.NODE_CLIENT_ROUTEFINDER) {
+                            player.pathToMoveClick(player.userPath, !Environment.NODE_CLIENT_ROUTEFINDER);
+
+                            if (!player.opcalled && player.hasWaypoints()) {
+                                player.processWalktrigger();
+                            }
                         }
                     }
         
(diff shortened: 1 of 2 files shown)


src/lostcity/engine/World.ts
```diff
@@ -691,11 +691,12 @@ class World {
                             player.pathToPathingTarget();
                             continue;
                         }
-        
-                        player.pathToMoveClick(player.userPath, !Environment.NODE_CLIENT_ROUTEFINDER);
-    
-                        if (!player.opcalled && player.hasWaypoints()) {
-                            player.processWalktrigger();
+                        if (!Environment.NODE_CLIENT_ROUTEFINDER) {
```
ultraviolet-jordan, line 694: This is probably breaking code. Test your code changes switching the env option on and off.

tannerdino: Tested interactions, movement, walktriggers, and following, with both client and server pathing. All seems to function properly, not sure if theres a better way to implement this to match the videos

ultraviolet-jordan: So why do you want this walktrigger here to run when you are only using server pathfinder and not client pathfinder? Then in the MoveClick packet you are running the walktrigger there both ways?

tannerdino: I assumed that when osrs transitioned to server pathfinding they moved the walktrigger from the packet handler to client input. But that mightve not been the case. Looks to me that osrs's server pathfinding is calculated in the packet handler and not in client input. This [video](https://youtu.be/TnBJZD9AcJE?t=9) from oct 2016 shows that spam clicking whilst frozen runs the walktrigger multiple times per tick.

I think sometime later in osrs they changed moved the server pathfinding to client input (along with walktriggers)? Need to set the waypoints in the packet handler in order for commands like p_walk and p_stopaction to work properly in the walktrigger. Which is why I use pathToMoveClick in the handler.

PR #963, opened 2024-10-27
feat(engine): npc and obj respawn scaling

Indio3:


src/lostcity/engine/World.ts
```diff
@@ -1344,9 +1345,10 @@ class World {
     removeObj(obj: Obj, duration: number): void {
         // printDebug(`[World] removeObj => name: ${ObjType.get(obj.type).name}, duration: ${duration}`);
         const zone: Zone = this.gameMap.getZone(obj.x, obj.z, obj.level);
+        const adjustedDuration = this.scaleByPlayerCount(duration);
         zone.removeObj(obj);
         obj.setLifeCycle(this.currentTick + duration);
```
ultraviolet-jordan, line 1350: You probably want to change this duration too.

src/lostcity/engine/World.ts
```diff
@@ -1571,6 +1573,13 @@ class World {
         }
         return this.players.next();
     }
+
+    scaleByPlayerCount(rate : number): number {
+        // not sure if it caps at 2k player count or not
+        const playerCount = Math.min(this.getTotalPlayers(), 2000);
+        return Math.trunc((((4000 - playerCount) * rate) / 4000)); // assuming scale works the same way as the runescript one
```
ultraviolet-jordan, line 1580: It computes faster if you do `return (((4000 - playerCount) * rate) / 4000) | 0` instead of `trunc`

src/lostcity/engine/World.ts
Indio3: good catch, will update

PR #973, opened 2024-11-01
feat(content): Family Crest quest

noted-shark:
- Add Family Crest quest
- Add Chaos Gauntlets and restore commented implementation of goldsmithing gauntlets

**Notes**

- This moves the `open_book` varp to 163, reassigning 149 to `crest_spells_levers_gauntlets`
  - Bits 0 through 3 contain the spells used on Chronozon
  - Bits 4 through 6 contain the levers set to the up position in the Witchaven dungeon
  - Bits 7 through 9 contain the player's choice of gauntlets (which is permanent)
- The structure of some of the gauntlet-related dialogue is a guess, as the ability to change gauntlets for 25k and Family Pest have impacted the OSRS & RS3 dialogue.
- There are some gates in Edgeville Dungeon gate leading to Chronozon. These gates were not present in RSC, were added around rev 194, and removed somewhere between 299 and 306. I can't find any footage of how they operated, but there are no opened versions available so I have assumed they close automatically after the player passes through.
- The [earliest reference](https://web.archive.org/web/20080130035157/https://www.rscommunity.de/index.php?option=content&task=view&id=233) I could find for the goldsmithing gauntlets xp boost is from May 2005, which mentions 56.5 xp per bar (a 2.5x boost). In RSC, the boost was 1.5x. I've implemented the 2.5x rate assuming the change was made when the quest was ported to RS2, but there's no evidence to back this up.
- Similarly, the chaos gauntlets were +1 max hit on RSC and +3 on RS2 ([source from Sep 2005](https://web.archive.org/web/20071029013850mp_/http://www.rscommunity.de/index.php?option=content&task=view&id=51)).
- For reference, crest_progress varp on OSRS: [map](https://mejrs.github.io/osrs.html?m=-1&z=1&p=0&x=3193&y=3350&varp=148), [detail](https://chisel.weirdgloop.org/varbs/display?varplayer=148)


data/src/scripts/general/scripts/invs.rs2
```diff
@@ -22,4 +22,9 @@ while ($i < enum_getoutputcount($slots)) {
     }
     $i = add($i, 1);
 }
-return(true);
\ No newline at end of file
+return(true);
+
+[proc,del_all_if_exists](inv $inv, namedobj $obj)
+if (inv_total($inv, $obj) > 0) {
+    inv_del($inv, $obj, inv_total($inv, $obj));
```
kylmp, line 29: maybe save inv_total to a var so you don't have to compute it twice

data/src/scripts/quests/quest_crest/configs/quest_crest.constant
```diff
@@ -0,0 +1,29 @@
+^crest_not_started = 0
+//@bla@str@I have agreed to restore Dimintheis' family crest to him.
+//@blu@I need to find his son @red@Caleb@blu@. The last Dimintheis heard of
+// him he was a great @red@chef@blu@ beyond @red@White Wolf Mountain
```
kylmp, line 4: need these comments still?

data/src/scripts/quests/quest_crest/scripts/crest_chronozon.rs2
```diff
@@ -0,0 +1,36 @@
+[ai_queue3,crest_chronozon]
+if (npc_findhero = true){
+    if (getbit_range(%crest_spells_levers_gauntlets, ^crest_wind_blast_used, ^crest_fire_blast_used) = calc(pow(2,4) - 1)) {
```
kylmp, line 3: I wonder if having the calc saved as a constant would be more useful/readable

data/src/scripts/quests/quest_crest/scripts/crest_johnathon.rs2
```diff
@@ -0,0 +1,133 @@
...
+    return;
+}
+if (last_useitem = 1doseantipoison |
+    last_useitem = 2doseantipoison |
+    last_useitem = 3doseantipoison |
+    last_useitem = 4doseantipoison |
+    last_useitem = 1dose2antipoison |
+    last_useitem = 2dose2antipoison |
+    last_useitem = 3dose2antipoison |
+    last_useitem = 4dose2antipoison) {
```
kylmp, line 26: Does strange fruit cure poison in 2004? 

Maybe there should be some sort of obj param for items that can cure poison

data/src/scripts/skill_mining/scripts/mine.rs2
```diff
@@ -141,7 +140,13 @@ if (random($chance) = ^true) {
     // stop mining, give ore and xp. Sudden says its sound_230
     sound_synth(sound_230, 0, 0);
     anim(null, 0);
-    inv_add(inv, db_getfield($data, mining_table:rock_output, 0), 1);
+    // The 'perfect' gold rocks in Family Crest are the same ID as regular gold rocks, so check to 
+    // give out the right ore if we're in that location. Only gold rocks are in this room.
```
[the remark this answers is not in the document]
tannerdino: Prospect was removed from osrs last year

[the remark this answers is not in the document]
kylmp: Oh wtf lol. I guess it makes sense who actually ever uses that

data/src/scripts/quests/quest_crest/scripts/crest_johnathon.rs2
noted-shark: Just checked on OSRS, using strange fruit post-quest gives the same message that trying to use antipoison does, so I think it should be added to the list of useable items here too:
image
(other items give the default "Nothing interesting happens" message)

I can't find any indication that the strange fruit was changed to cure poison, so it's probably reasonable to assume it always did.

data/src/scripts/skill_mining/scripts/mine.rs2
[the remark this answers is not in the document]
noted-shark: I couldn't find any footage of someone prospecting the rocks. The "You manage to mine some gold" message is unchanged, so my guess is that they just changed the output item and none of the messaging but it's hard to know for sure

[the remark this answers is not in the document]
kylmp: Yea sorry didnt mean to send you on a goose chase for that, unlikely someone recorded prospecting those

data/src/scripts/general/scripts/invs.rs2
noted-shark: Yep, good call

data/src/scripts/quests/quest_crest/configs/quest_crest.constant
noted-shark: I left them there in case we get to the point of implementing quest journal text for each step later down the track. Happy to pull it out though, having only one step with the journal text saved isn't super useful

kylmp (approved):

tannerdino:
I searched through the fansites I've saved, and the oldest evidence I could find of goldsmith gauntlets being 2.5x instead of 1.5x is from this [forum post](https://www.rsbandb.com/forums/viewtopic.php?f=2&t=4728.html) dated March 9, 2005. Couldnt find anything post release about 1.5x

Imo, I think we should have it at 2.5x. I will keep searching for earlier evidence

tannerdino:
> Found an August 2004 mention of the 2.5x rate in the second reply in this thread: https://web.archive.org/web/20040820223401/http://runevillage.net/ThePub/viewtopic.php?p=1257194&amp

Nice find, it's official then

PR #976, opened 2024-11-02
fix(content): add missing seaslug rewards

hex-agon:
The seaslug quest was missing it's rewards. As far as i could find these rewards were present in 2004.

https://web.archive.org/web/20040722192346/http://www.tip.it:80/runescape/index.php?quest_id=60
https://web.archive.org/web/20041230201616/http://www.runehq.com/cacheguides/viewquestguide00180.htm


Indio3:
the seaslug quest is not missing it's rewards: https://github.com/2004Scape/Server/blob/main/data/src/scripts/areas/area_ardougne_east/scripts/caroline.rs2#L42

PR #978, opened 2024-11-02
fix(engine): fixes #974

### LINKED ISSUE #974: Movement bug with clicking pathing entities
cocknballs

- I have confirmed the new pathfinder did not cause this, probably caused by recent changes to moving related code.
- Occurs when using server or client pathfinder.
- Does not seem to occur with Locs or Objs.
- Can confirm this is a new bug.

tannerdino:
movement related bug:

https://github.com/user-attachments/assets/26656573-f53f-43c8-864f-ec68c81ae7f5

follow unaffected:

https://github.com/user-attachments/assets/7ea107ce-7ee1-486d-bc2f-d545a76f0ad2

### DIFF
--- a/src/lostcity/engine/World.ts
+++ b/src/lostcity/engine/World.ts
@@ -672,6 +672,7 @@ class World {
                 }
 
                 if (isNetworkPlayer(player) && player.decodeIn()) {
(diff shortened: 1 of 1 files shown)


ultraviolet-jordan:
Fixes #974

PR #981, opened 2024-11-04
fix(content): bronze bar superheating

noted-shark:
- Ensure both primary and secondary ore are present, fixing bug where bronze bars could be superheated with only tin ore
- Add OSRS failure message if both copper and tin not present image


tannerdino:
Good work!

Issue #982, opened 2024-11-04
Inaccurate player routefinder to overlapping target


tannerdino:
Currently we use osrs's implementation. Early osrs and pre-2007 runescape, players and npc's handled routefinding to overlapping targets, the exact same way: Moving to a random direction

Example 1: May 11, 2006 [video](https://youtu.be/VzOz2FuXn3k&t=106)

https://github.com/user-attachments/assets/2ef2602a-3b01-434a-8975-6b0da619d179

Example 2: June 29, 2007 [video](https://youtu.be/MfYoyM8EVBs?t=78)

https://github.com/user-attachments/assets/08407718-f051-436b-930a-f8b24204fca2

ultraviolet-jordan:
lots of unknowns about this without being able to test on an original osrs server or something but this can be cheaply done with something like this:

![image](https://github.com/user-attachments/assets/be994ee4-0f4c-45ab-b4a1-6254222591b2)

PR #983, opened 2024-11-05
fix(engine): players now naive path if target is ontop of them

tannerdino:
Fixes https://github.com/2004Scape/Server/issues/982
Thanks to https://github.com/2004Scape/Server/issues/982#issuecomment-2455896662

However a new issue arises:

https://github.com/user-attachments/assets/9ce12f1c-50e3-40b9-84bf-1f9d4ed4baef

In order to fix this I added the same work around to pathToMoveClick():

https://github.com/tannerdino/Server/blob/72cbca79c438eb37097d9cf6fe0037247a90604e/src/lostcity/entity/PathingEntity.ts#L403-L408

### DIFF
--- a/src/lostcity/entity/PathingEntity.ts
+++ b/src/lostcity/entity/PathingEntity.ts
@@ -402,6 +402,10 @@ export default abstract class PathingEntity extends Entity {
 
     pathToMoveClick(input: number[], needsfinding: boolean): void {
         if (this.moveStrategy === MoveStrategy.SMART) {
+            if (this.target && this.target instanceof PathingEntity && Environment.NODE_CLIENT_ROUTEFINDER && CoordGrid.intersects(this.x, this.z, this.width, this.length, this.target.x, this.target.z, this.target.width, this.target.length)) {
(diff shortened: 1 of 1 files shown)


src/lostcity/entity/PathingEntity.ts
```diff
@@ -402,6 +402,10 @@ export default abstract class PathingEntity extends Entity {
 
     pathToMoveClick(input: number[], needsfinding: boolean): void {
         if (this.moveStrategy === MoveStrategy.SMART) {
+            if (this.target && this.target instanceof PathingEntity && Environment.NODE_CLIENT_ROUTEFINDER && CoordGrid.intersects(this.x, this.z, this.width, this.length, this.target.x, this.target.z, this.target.width, this.target.length)) {
+                this.queueWaypoints(findNaivePath(this.level, this.x, this.z, this.target.x, this.target.z, this.width, this.length, this.target.width, this.target.length, 0, CollisionType.NORMAL));
+                return;
+            }
```
tannerdino, line 408: I dont know if I like this fix though, seems like the movement code in playerProcessSetup needs to be rewritten with  accommodation of new findings (yellow-x walktrigger, and naive path for targets ontop of players)

PR #994, opened 2024-11-12
feat(engine): ai_spawn triggers, 0 xp spawnhitting bug, and process order of npc hunt's

tannerdino:
0 xp spawnhitting:
- [Video](https://youtu.be/yYRSpv2AGq4) by keyboardOSRS (patched)
- Ours with proposed fixes:

https://github.com/user-attachments/assets/2d02e84e-6037-4643-8da1-3399ae20d12c

- ash tweet:

image

Respawn is moved to processNpcs() because:
- Npc's with hunt rates of 1, will always hunt 1 tick after *visibly* respawning (which is a total of 2 ticks after being added to the world). If respawns were processed before hunts, npc's with hunt rates of 1 would hunt the same tick it *visibly* respawns

https://github.com/user-attachments/assets/3f365058-4640-4d29-91e1-7322dcb7b4ea

https://github.com/user-attachments/assets/7ab173c3-da5e-47d2-8a84-4f0c0714ab77

- ai_spawn scripts take a tick to process in early osrs. This is the mechanism, along with pvm bonus xp handled with a varn, that allows 0 xp spawn hitting to work.

Obj/npc/loc hunting was moved to processNpcs() because:
- In osrs, npc's will only hunt *players* in the processWorld() part of the tick. This video shows that the npc will hunt the obj the same tick it is dropped (meaning its processed after processPlayerSetup()):

https://github.com/user-attachments/assets/849fec13-f5f6-4db9-91e1-366c9fddf828
- I'm pretty certain that at gwd the npc's will still hunt you if you unequip a god item in the same tick. Need a video of this though

### DIFF
--- a/src/lostcity/entity/Npc.ts
+++ b/src/lostcity/entity/Npc.ts
@@ -51,8 +51,8 @@ export default class Npc extends PathingEntity {
     baseLevels: Uint8Array = new Uint8Array(6);
 
     // runtime variables
-    vars: Int32Array;
-    varsString: string[];
+    readonly vars: Int32Array;
+    readonly varsString: string[];
 
     // script variables
     activeScript: ScriptState | null = null;
@@ -159,8 +159,8 @@ export default class Npc extends PathingEntity {
             }
             this.resetHeroPoints();
             this.queue.clear();
-            this.vars = new Int32Array(VarNpcType.count);
-            this.varsString = new Array(VarNpcType.count);
+            this.vars.fill(0);
+            this.varsString.fill('');
             this.defaultMode();
 
             const npcType: NpcType = NpcType.get(this.type);
--- a/src/lostcity/entity/Npc.ts
+++ b/src/lostcity/entity/Npc.ts
@@ -52,7 +52,7 @@ export default class Npc extends PathingEntity {
 
     // runtime variables
     readonly vars: Int32Array;
-    readonly varsString: string[];
+    readonly varsString: (string | undefined)[];
 
     // script variables
     activeScript: ScriptState | null = null;
@@ -160,7 +160,7 @@ export default class Npc extends PathingEntity {
             this.resetHeroPoints();
             this.queue.clear();
             this.vars.fill(0);
-            this.varsString.fill('');
+            this.varsString.fill(undefined);
             this.defaultMode();
 
             const npcType: NpcType = NpcType.get(this.type);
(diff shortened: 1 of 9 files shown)


src/lostcity/entity/Npc.ts
```diff
@@ -159,6 +159,8 @@ export default class Npc extends PathingEntity {
             }
             this.resetHeroPoints();
             this.queue.clear();
+            this.vars = new Int32Array(VarNpcType.count);
```
ultraviolet-jordan, line 162: do this instead:

`this.vars.fill(0);`
`this.varsString.fill(undefined);`

idk why they aren't but these should be `readonly`

```ts
    readonly vars: Int32Array;
    readonly varsString: string[];
 ```

Comment (tannerdino) [reply src/lostcity/entity/Npc.ts]: Thank you! I ended up doing this because undefined wouldn't work:
`this.vars.fill(0);`
`this.varsString.fill('');`

Comment (ultraviolet-jordan) [reply src/lostcity/entity/Npc.ts]: Probably bcuz `varsString` isnt typed properly. `this.varsString.fill('');` isn't right either.

Comment (ultraviolet-jordan) [reply src/lostcity/entity/Npc.ts]: Yeah Im pretty sure its `this.varsString.fill(undefined);`

Comment (Pazaz) [reply src/lostcity/entity/Npc.ts]: If it were me, I would do something like varsString.splice(0) and leave it as string[]

example:
```ts
const varsString: string[] = [];

for (let i = 0; i < 1000; i++) {
    varsString.push(i.toString());
}

varsString.splice(0); // varsString is back to []
```

ultraviolet-jordan: > If it were me, I would do something like varsString.splice(0) and leave it as string[]
> 
> example:
> 
> ```ts
> const varsString: string[] = [];
> 
> for (let i = 0; i < 1000; i++) {
>     varsString.push(i.toString());
> }
> 
> varsString.splice(0); // varsString is back to []
> ```

image

Pazaz: I think explicitly having undefined there is overly verbose. It's not in the strict language rules for a reason. In JS/TS it's also not the array itself storing undefined, but the *array access* returning undefined.

If you're really serious about it @ultraviolet-jordan we should turn on `noUncheckedIndexedAccess: true` in tsconfig.
https://github.com/microsoft/TypeScript/issues/49169
https://github.com/microsoft/TypeScript/pull/39560

ultraviolet-jordan: imo indexing out of bounds in an array should be illegal and should crash the program

Pazaz: lai6xuz2n6ia1

PR #1015, opened 2024-11-21
fix(engine): walktrigger env settings and moveClickRequest changes

tannerdino:
- adds ~playerwalk3 proc
- adds NODE_... setting for where walktrigger is ran in the world cycle. 0 = processed in packet handler. 1 = processed in player setup (client input). 2 = processed in player movement. For setting 0, path is set in the packet handler. For settings 1 & 2, path is set in player setup.
- defaults both path to be set and walktrigger to be processed immediately in the packet handler. In the span of OSRS's lifetime, the location of where walktrigger was processed for yellowx was changed from: packet handler -> client input -> player movement. Videos: [May 4, 2013 OSRS](https://youtu.be/dKfFkEyqD8E?t=9) (before server pathing update), [Oct 17, 2016 OSRS](https://youtu.be/TnBJZD9AcJE) (after server pathing update)
- p_op and p_walk should not set moveClickRequest, and moveClickRequest is reset on login. If a queue is pending, you will not move when placing the cannon, unless moveClickRequest is false. In osrs on login, without moving, you will walk with the cannon just fine. But once you yellowx, your next cannon + queue will prevent movement for the length of the delay.

Cannon + pending queue (notice how I stop moving as soon as I turn on prayers):

https://github.com/user-attachments/assets/9baedf56-d642-4e91-92e7-f69bbf669ad4

https://github.com/user-attachments/assets/2c69dec8-0f87-4cc6-9a6b-5eb65251df34

Cannon + pending queue after logout:

https://github.com/user-attachments/assets/ff1949c1-0120-465e-bd93-d8508c60679e

https://github.com/user-attachments/assets/dbc5cd6d-dc74-4790-8c24-ce011565109e

- set moveClickRequest to false on opheld and opheldu null interactions, to match early osrs bug abuse videos such as: https://youtu.be/GPfN3Isl2rM (smuggles games room effect, which clears on engine queue). This could theoretically work with other items like books (as long as it doesnt clear movement) - todo: find proof of items that work *other than* dueling ring as shown in the linked video.

- only remove the flag on the map for op packet errors. Null obj example:

https://github.com/user-attachments/assets/787c54eb-768b-4a1e-888f-1b23fa82a43b

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -21,6 +21,7 @@
 # NODE_STAFF=pazaz
 # NODE_CLIENT_ROUTEFINDER=true
 # NODE_SOCKET_TIMEOUT=true
+# NODE_WALKTRIGGER_SETTING=0
 
 ## login server
 # LOGIN_HOST=localhost
--- a/data/src/scripts/player/playerwalk.rs2
+++ b/data/src/scripts/player/playerwalk.rs2
@@ -1,0 +1,8 @@
+// https://x.com/JagexAsh/status/1605130887292751873
+[proc,playerwalk3](coord $coord)
+p_walk($coord);
+if (%player_run = ^player_run_off) {
+    p_delay(sub(distance(coord, $coord), 1));
+} else {
+    p_delay(sub(divide(distance(coord, $coord), 2), 1));
+}
--- a/data/src/scripts/quests/quest_mcannon/scripts/cannon.rs2
+++ b/data/src/scripts/quests/quest_mcannon/scripts/cannon.rs2
@@ -82,14 +82,16 @@ if ($setupcoord = null) {
     mes("There isn't enough space to set up here.");
     return;
 }
-~forcewalk2($setupcoord);
+~playerwalk3($setupcoord);
+
+facesquare($center);
+p_delay(0);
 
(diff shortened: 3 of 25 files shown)


tannerdino:
Is NODE_WALKTRIGGER_SETTING in .env the right approach here? We'll probably never have it set to anything other than 0, since Lost City, the project, will probably not do OSRS revs. But it could be nice for server owners who want to match certain OSRS updates?

If the goal of this engine is to be very general with the ability to match different eras with env settings, then I guess a NODE_WALKTRIGGER_SETTING is good

Pazaz:
> Is NODE_WALKTRIGGER_SETTING in .env the right approach here? 

I think it's a fine approach, basically the same as a config file and we can pre-make template .env files for future eras.

My only concern, if we have it as `SETTING=version`... what if we discover a "new version" change in-between? I'm thinking generally since we would want to keep the same convention for other feature flags.  
Adding to the end could be confusing, and shifting everything might break people's setups. Keep it as-is for now but keep that in mind... maybe I should come up with a tryParseEnum for this case.

ultraviolet-jordan:
if ur doing this then u can delete this:

```ts
    clearWaypoints(): void {
        super.clearWaypoints();
    }
```

PR #1016, opened 2024-11-21
refactor(engine): create heropoints class & improve chatmodes

ultraviolet-jordan:
- Remove heropoints duplicate code by creating a class for it.
- Improve player chatmodes code.
- Some imports organizing.

### DIFF
--- a/src/lostcity/entity/Player.ts
+++ b/src/lostcity/entity/Player.ts
@@ -12,6 +12,8 @@ import ObjType from '#lostcity/cache/config/ObjType.js';
 import ScriptVarType from '#lostcity/cache/config/ScriptVarType.js';
 import SeqType from '#lostcity/cache/config/SeqType.js';
 import VarPlayerType from '#lostcity/cache/config/VarPlayerType.js';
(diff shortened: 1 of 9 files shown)


src/lostcity/entity/Player.ts
```diff
@@ -238,15 +234,9 @@ export default class Player extends PathingEntity {
...
-    chatModes: {
-        publicChat: ChatModePublic;
-        privateChat: ChatModePrivate;
-        tradeDuel: ChatModeTradeDuel;
-    } = {
-            publicChat: ChatModePublic.ON,
-            privateChat: ChatModePrivate.ON,
-            tradeDuel: ChatModeTradeDuel.ON,
-        };
+    publicChat: ChatModePublic = ChatModePublic.ON;
```
Jameskmonger, line 237: not sure this is an improvement, they are closely related so imo it makes sense to group them

ultraviolet-jordan: it was ugly and now it is handsome

PR #1017, opened 2024-11-24
fix(content): mage arena dialogues

kylmp:
https://youtu.be/mQmeH32PdNk?t=19

This dialogue confirms they are not using "Gielinor" in their dialogue, changed other usages of the word to "RuneScape" off an assumption it is the same.

Also NPC say texts were incorrect compared to the video

### DIFF
--- a/data/src/scripts/areas/area_mage_arena/scripts/kolodion.rs2
+++ b/data/src/scripts/areas/area_mage_arena/scripts/kolodion.rs2
@@ -9,7 +9,7 @@ if (%magearena < ^mage_arena_started) {
         ~chatnpc("<p,angry>Hah! A wizard of your level? Don't be absurd.");
         return;
     }
-    ~chatnpc("<p,shifty>I am the great Kolodion, master of battle magic, and|this is my battle arena. Top wizards travel from all over|Gielinor to fight here.");
+    ~chatnpc("<p,shifty>I am the great Kolodion, master of battle magic, and|this is my battle arena. Top wizards travel from all over|RuneScape to fight here.");
     @multi3("Can I fight here?", kolodion_can_i_fight_here, "What's the point of that?", kolodion_whats_the_point, "That's barbaric!", kolodion_barbaric);
 }
 if (~mage_arena_in_progress = true) {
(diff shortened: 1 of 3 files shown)


tannerdino:
The "Let us continue with our battle" is when you leave the fight early and return later! The real npc_say's are in kolodion_fight.rs2

data/src/scripts/areas/area_mage_arena/scripts/kolodion.rs2
```diff
@@ -85,7 +85,7 @@ if ($choice = 1) {
 [label,kolodion_yes_indeedy]
 ~chatplayer("<p,happy>Yes indeedy.");
 ~chatnpc("<p,happy>Good, good. You have a healthy sense of competition.");
-~chatnpc("<p,neutral>Remember, traveller - in my arena, hand-to-hand|combat is useless. Your strength will diminish as you|enter the arena, but the spells you can learn are|amongst the most powerful in all of Gielinor.");
+~chatnpc("<p,neutral>Remember, traveller - in my arena, hand-to-hand combat|is useless. Your strength will diminish as you enter the|arena, but the spells you can learn are amongst the|most powerful in RuneScape.");
```
tannerdino, line 88: In the video it looks like the dialogue should be `~chatnpc("<p,neutral>Remember traveller, in my arena hand to had combat|is useless.  (double space?) Your strength will diminish as you enter the|arena, but the spells you can learn are amongst the most powerful in RuneScape.")`

kylmp: image

Ah yea, no dashes in hand-to-hand, updated

tannerdino: Nevermind it looks like 1 space matches up with that screenshot, before "Your strength"

image

Also looks like `Good, good, you have a healthy sense of competition.` instead of our:
https://github.com/2004Scape/Server/blob/f704f509df7cb8237ebe698a8793132dd9dfd9c4/data/src/scripts/areas/area_mage_arena/scripts/kolodion.rs2#L87

image

image

`Ok, let's fight.` instead of our `Okay, let's fight.`

image

`Not just any magician can enter, only the most|powerful, the most feared. Before you can use the|power of this arena, you must prove yourself against|me.` 
instead of:
 `Not just any magician can enter - only the most|powerful and most feared. Before you can use the|power of this arena, you must prove yourself against|me.`

PR #1030, opened 2024-12-31
fix(content): Oziach should give spare key to maze during quest if lost

cernodile:
References:
- https://forum.tip.it/topic/75433-melzars-maze-not-an-option/ 
- https://forum.tip.it/topic/42769-melzars-maze-key/

While RSC Dragon Slayer had you trade other players or buy a new key from Legends Guild, with RS2, Maze key can not be traded - locking f2p out if key is lost.
This also seems to line up with known varp values for Dragon Slayer in OSRS (even though the quest was reworked in 2007): https://chisel.weirdgloop.org/varbs/display?varplayer=176

I ran a test on local server by setting dragon_progress varp to 1 and talking to Oziach, trying to do the drop trick. I found nothing obvious that had broken as result of this, but a review is appreciated in any case.


dginovker:
Very sexy

PR #1054, opened 2025-01-06
fix(engine): clear multiple stat_advance engine queues of one skill type

tannerdino:
in osrs, only the last level of a stat level up is shown per skill.

https://github.com/user-attachments/assets/a852d881-b5cb-4c82-8893-e0bafe80286b

### DIFF
--- a/src/engine/entity/Player.ts
+++ b/src/engine/entity/Player.ts
@@ -608,6 +608,28 @@ export default class Player extends PathingEntity {
         }
     }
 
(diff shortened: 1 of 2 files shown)


tannerdino:
Also an ash [tweet](https://x.com/JagexAsh/status/1525156067960934401)

image

Pazaz:
Good enough for me! If we need to revisit this we will later. QoL incoming...

PR #1056, opened 2025-01-07
fix(content): Fix Demon Slayer issues

Gregf36665:
# Changes
- Remove magic numbers and use constants for Demon Slayer
- Allow reclaiming of Silverlight both during the quest (free) and after (500gp)
- Prevent weakened Delirth from attacking you since this closes the interface ofr the incantation
- Allow reclaiming of keys when needed
- Prevent players from hiding items in banks to get extra items
- Remove obsolete `set_demon_progress` and `update_demon_progress`
- Allow only players who are meant to be fighting Delirth to fight. Previously someone could drop trade Silverlight and jump to the end of the quest
- Fix `getvar` and `setvar` to need only 1/2 args rather than 2/3 and ignoring the first one.
- Fix bug where quest could be restarted by talking to the gypsy through multiple paths
- Update gypsy dialogue to reference being old and legs in the blue moon inn

# Sources
Since this quest was changed it's slightly more challenging to piece it together.

- https://www.youtube.com/watch?v=sGMYNUNMDT4 Prove it's 500gp to buy another one

- https://www.youtube.com/watch?v=0v5ufPbK0jw best source for historic dialogue

- https://2007rshelp.com/quests.php?id=10 source for reclaim before end of quest is free

- https://chisel.weirdgloop.org/varbs/display?varplayer=222 varbits have changed (for OSRS) now compared to back then so it is hard to know what stage should be what. It references sword cases that don't exist yet.

- https://forum.tip.it/profile/140266-trobigod/content/page/5/ gypsy is old
- https://classic.runescape.wiki/w/Transcript:Gypsy gypsy is this old!

Runescape wiki for one month said you had to get all the keys again https://runescape.wiki/w/Silverlight?oldid=41265 (which you do if you don't claim it the first time).

Guides from the era (and the wiki within 1 month) say that you can reclaim the sword as long as you have handed in the keys at least once.

Since Siegfried Erkle is not implemented I have not added buying the sword from him for 75gp. This was removed in 29 August 2006 ([Update](https://runescape.wiki/w/Silverlight#Update_history):


tannerdino:
I did some digging of my own on the silverlight, the earliest confirmation of being able to buy it I could find was from Feb 2005 from this [forum post](https://www.neoseeker.com/forums/2410/t500458-did-get-holiday-drop/2.htm).

image

This is before shadow of the storm, which was before the running theory at the time of when the silver light was changed to be buy-backable

ultraviolet-jordan:
https://github.com/user-attachments/assets/1b48779f-51a6-44b9-8557-2921957ee0e3

data/src/scripts/quests/quest_demon/scripts/demon_slayer.rs2
```diff
@@ -1,39 +1,29 @@
...
 ~send_quest_complete(questlist:demon, silverlight, 250, ^demon_questpoints, "You have completed the\\nDemon Slayer Quest!");
 
 // Search Drain
 [oploc1,varrock_palace_sewer_drain]
-if (%demon_progress < 2) {
+if (%demon_progress < ^demon_key_hunt) {
     ~mesbox("I can see a key but I can't quite reach it...");
 } else {
+    if (~demon_has_prysin_key = true){
+        // Unsure if this is the authentic message but you should not get a message about a key if you have it already
```
ultraviolet-jordan, line 14: You could be like:

```ts
if (%demon_progress < ^demon_key_hunt) {
    ~mesbox("I can see a key but I can't quite reach it...");
} else if (~demon_has_prysin_key = false) {
    ~mesbox("This is the drainpipe running from the kitchen sink to the sewer.");
    ~mesbox("I can see a key just inside the drain.|That must be the one Sir Prysin dropped.");
    ~mesbox("I don't seem to be able to reach it. I wonder if I can dislodge it somehow and knock it down into the sewer.");
}
```

and then don't need a guessed message.

Gregf36665: In fact I don't think there should be any restrictions on flushing it out:

> This is the only Silverlight key which may be obtained both before and after the player is progressing through the Demon Slayer quest
[source](https://classic.runescape.wiki/w/Silverlight_key_3)
No one (osrs/rs3 historic) actually claims that you cannot get the key or that you can't get more than 1 key.

Guess I've got to update it that way!

tannerdino: Couldve easily been changed on rs2 conversion

Gregf36665: For now I've put it as able to happen at anytime since I could not find a source saying differently in 2004.

PR #1062, opened 2025-01-07
fix(content): fix typo, improve quest stage setting, also fix quest-related black knight aggression

### LINKED ISSUE #1058: Possible issues with Black Knights Fortress
First off, a possible typo:

2025-01-07-222003_792x534_scrot

Second, I took the mission, found out about the secret weapon, then put the cabbage in the hole. A few minutes later, I disconnected and died. I walked back to Sir Amik Varse, and he doesn't recognize that I've destroyed the secret weapon.

KeepBotting:
Sorry for the double PR, I had to quickly relearn `git remote` and branch from the correct place.

This closes #1058 with the added bonus that I _think_ I have fixed black knight overhead text/aggression correctly (tested both areas several times and saw no issues with removing the `$has_spoken` stuff).

### DIFF
--- a/data/src/scripts/quests/quest_blackknight/scripts/quest_blackknight.rs2
+++ b/data/src/scripts/quests/quest_blackknight/scripts/quest_blackknight.rs2
@@ -21,9 +21,9 @@ if ($is_outside = true & ((inv_getobj(worn, ^wearpos_hat) ! bronze_med_helm) | (
 [oploc1,blackknight_door_to_grill]
 def_boolean $is_outside = ~check_axis(coord, loc_coord, loc_angle);
 if ($is_outside = false) {
-    def_boolean $has_spoken = false;
-    $has_spoken = ~set_black_knight_aggro(1_47_54_17_55, $has_spoken);
-    ~set_black_knight_aggro(1_47_54_15_55, $has_spoken);
+    // black knights guarding the ladder down to the grill will aggress the player
+    ~set_black_knight_aggro(1_47_54_17_55);
+    ~set_black_knight_aggro(1_47_54_15_55);
 }
 ~open_and_close_door(loc_param(next_loc_stage), $is_outside, false);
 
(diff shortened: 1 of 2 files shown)


data/src/scripts/quests/quest_blackknight/scripts/quest_blackknight.rs2
```diff
@@ -89,11 +88,10 @@ if ($option = 1) {
 } else {
    ~chatplayer("<p,neutral>I don't care. I'm going in anyway.");
    ~open_and_close_door(loc_param(next_loc_stage), $is_inside, false);
-   // todo: Set all black knights in the banquet hall to attack player and yell "Die intruder!"
-   def_boolean $has_spoken = false;
-   $has_spoken =  ~set_black_knight_aggro(0_47_54_13_59, $has_spoken);
-   ~set_black_knight_aggro(0_47_54_21_59, $has_spoken);
-   ~set_black_knight_aggro(0_47_54_22_55, $has_spoken);
+   // black knights in the banquet hall will aggress the player
+   ~set_black_knight_aggro(0_47_54_13_59);
```
Indio3, line 92: Don't have to do this but this can be simplified to use to npc_huntall requiring los instead (you can check temple of ikov code for aggroing armadyl guardians as reference)

data/src/scripts/quests/quest_blackknight/scripts/quest_blackknight.rs2
```diff
@@ -153,14 +154,11 @@ if (%blackknight_progress = 1) {
     mes("I can't hear much right now.");
 }
 
-[proc,set_black_knight_aggro](coord $coord, boolean $has_spoken)(boolean)
+[proc,set_black_knight_aggro](coord $coord)(boolean)
```
Pazaz, line 157: Need to remove the return from this proc since its usage changed :)

Get rid of the `(boolean)` at the end of this line and the `return` statement at the bottom of the proc.

PR #1065, opened 2025-01-08
fix(engine): in packets failed accounting towards queue limit

ultraviolet-jordan:
For example, the current logic, makes you not able to spam click to pickup items on the floor, because your "queue" essentially gets filled with handles to objs that no longer exist.

Now it will only count towards the limiter when they are successfully handled.

idk 216


Pazaz:
Better behavior for sure, but this lets bad actors send any # of malicious packets to run in 1 tick as long as it fits in the socket buffer (65KB). Maybe unsuccessful packets can increment clientLimit instead of userLimit so clients all share the same max processing time?

```ts
if (success && message.category === ClientProtCategory.USER_EVENT) {
    userLimit++;
} else {
    clientLimit++;
}
```

PR #1073, opened 2025-01-09
feat(content): Add drops to tribesman

Gregf36665:
# Changes
- Added drop table for tribesman

# Sources
- https://oldschool.runescape.wiki/w/Tribesman
- https://runescape.wiki/w/Wandering_warrior
- https://classic.runescape.wiki/w/Tribesman

OSRS has trading sticks and cleaning cloths. They happen to align, weight and value wise with some RSC drops so 138 as the base rate was used (from OSRS and RS3) but later drops replaced with their RSC drops.

<!-- #Outside of scope ->

### DIFF
--- a/data/src/scripts/drop tables/scripts/tribesman.rs2
+++ b/data/src/scripts/drop tables/scripts/tribesman.rs2
@@ -1,4 +1,4 @@
-[ai_queue3,tribalman]
+[ai_queue3,tribesman]
 gosub(npc_death);
 if (npc_findhero = false){
     return;
--- a/data/src/scripts/drop tables/scripts/tribesman.rs2
+++ b/data/src/scripts/drop tables/scripts/tribesman.rs2
@@ -1,4 +1,4 @@
-[ai_queue3,tribesman]
+[ai_queue3,tribalman]
 gosub(npc_death);
 if (npc_findhero = false){
     return;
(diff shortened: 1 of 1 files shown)


data/src/scripts/drop tables/scripts/tribesman.rs2
```diff
@@ -0,0 +1,56 @@
...
+    obj_add(npc_coord, gold_ore, 1, ^lootdrop_duration);
+} else if ($random < 119) {
+    obj_add(npc_coord, 2dose2antipoison, 1, ^lootdrop_duration);
+} else if ($random < 120) {
+    obj_add(npc_coord, 3dose2antipoison, 1, ^lootdrop_duration);
+} else if ($random < 121) {
+    obj_add(npc_coord, bread, 1, ^lootdrop_duration);
+} else if ($random < 122) {
+    obj_add(npc_coord, tin_ore, 1, ^lootdrop_duration);
+} else if ($random < 123) {
```
Pazaz, line 54: Since the table ends at 123 here, 138-123 leaves a 15/128 chance of nothing. Just something to note if it's not intended

Gregf36665: I missed the herb (11/138) and RDT (2/138) which would give (2/138) for nothing

PR #1078, opened 2025-01-10
fix(content): party room fm, enchanted jewellry, stall stealing fixes

tannerdino:
- The previous firemaking check at the party room used this [blog post](https://oldschool.runescape.wiki/w/Update:Spirit_trees,_clues_%26_more!) as a source. `Cannons can no longer be setup within the Party room and can now be placed where the Party room used to be.` This [video](https://www.youtube.com/watch?v=BmL5W1CMvGc) shows you can light fires in the party room.

VS--YouTube-Kingys99woodcuttingparty-1’05”

- Adds ring of life 'bug', teleporting at 0 hp. From this [video](https://youtu.be/LSpyUVXa4LE?t=200):

https://github.com/user-attachments/assets/ef609ad1-cb6d-4294-b943-ebc0b42dc4f3

- Refactors a bit of the enchanted jewellry code

- Refactors stealing code: Replace now redundant procs with npc_huntall, remove unnecessary 'stolen_from' vars, and change stolen from mechanics to longqueue, to match osrs (stealing from stall recently stops ur movement during p_delays until the longqueue is over)

- Matches tutorial island bone with the regular bone delay

- Adds the alleged watchman spawn in f2p bug. Mining is similar - you don't even need a pickaxe to get a random event lol. Watchman doesn't drop any members items other than herbs (on p2p worlds)

- Fixes pizza toppings giving crafting xp instead of cooking xp


tannerdino:
I wonder what happens if you light a log on every tile in the party room...

data/src/scripts/general/scripts/enchanted_jewellry/ring_of_life.rs2
```diff
@@ -0,0 +1,11 @@
...
+if (p_finduid(uid) = true) {
+    if (inv_getobj(worn, ^wearpos_ring) ! ring_of_life | ~wilderness_level(coord) > 30 | ~pre_tele_checks(coord) = false) {
+        return;
+    }
+    if (stat(hitpoints) <= scale(10, 100, stat_base(hitpoints))) {
+        mes("Your Ring of Life saves you and is destroyed in the process.");
+        inv_delslot(worn, ^wearpos_ring);
+        ~player_teleport_normal(map_findsquare(0_50_50_21_18, 0, 2, ^map_findsquare_lineofwalk));
+    }
+}
```
ultraviolet-jordan, line 11: Should there be a false condition?

tannerdino: Acts as a normal teleport, with the `~pre_tele_checks()` proc. So you wouldnt be able to leave a duel/minigame etc. Main thing changed with ring of life is putting the ring of life check before death.


https://github.com/2004Scape/Server/blob/1b1632d7d775f38e678819d87c2ac1e9251e6f65/data/src/scripts/player/scripts/damage.rs2#L16-L21

Guess it could be like:
```
if (stat(hitpoints) = 0){
    queue(player_death, 0);
}

~ring_of_life_check;
```

Which is probably more likely? Difference with my current code, being that if u had something queue on you during the ring of life teleport, that'd be processed before player_death

PR #1081, opened 2025-01-10
fix(content): add queen's guard to ice_warrior droptable

KeepBotting:
I verified this against RSC/OSRS wikis and also checked in-game in OSRS, these warriors should be on the drop table too!


KeepBotting:
Just realized this prevents regular ice warriors not only from dropping loot, but from even dying at all :rofl:

data/src/scripts/drop tables/scripts/ice_warrior.rs2
```diff
@@ -1,4 +1,7 @@
-[ai_queue3,ice_warrior]
+[ai_queue3,ice_warrior] @ice_warrior_drop_table;
```
Pazaz, line 1: You can add a category for both NPCs and make this like `ai_queue3,_ice_warrior`, rather than a shared label

PR #1082, opened 2025-01-11
fix(content): cooking-pizza-bugfix

xVye:
Fixed a bug where making pizzas would result in giving crafting instead of cooking xp.


Gregf36665:
@tannerdino put this fix into #1078

Pazaz:
> @tannerdino put this fix into #1078

Merging this one to recognize a new contributor and encourage smaller-scope fixes :)

PR #1088, opened 2025-01-11
fix(content): allow 1 bait fishing

Gregf36665:
# Changes
- Allow fishing with feathers to continue if the last fish will replace your last feather.
- Allow fishing with bait to continue if the last fish will replace your last bait.

# Sources
- None besides memories in discord


cernodile:
Does bait fishing have same logic for if 1 bait is only left? Might make sense to add same check there.

Gregf36665:
> Does bait fishing have same logic for if 1 bait is only left? Might make sense to add same check there.

Added it in

PR #1104, opened 2025-01-16
fix(content): alpha 2.0 login mes

KeepBotting:
I've encountered several people who weren't aware of this so I think we should add it back in. Better late than never!


ultraviolet-jordan:
Maybe mention how there are no resets after?

PR #1107, opened 2025-01-18
fix(engine): Close modal improvements

henke96:
Fixes:
- Handle close modal request at start of player processing
- Make last login info modal purely client sided

### DIFF
--- a/src/engine/entity/Player.ts
+++ b/src/engine/entity/Player.ts
@@ -272,7 +272,7 @@ export default class Player extends PathingEntity {
     cameraPackets: LinkList<CameraInfo> = new LinkList();
     timers: Map<number, EntityTimer> = new Map();
     tabs: number[] = new Array(14).fill(-1);
-    modalState = 0; // 1 - main, 2 - chat, 4 - side, 8 - tutorial, 16 - welcome
+    modalState = 0; // 1 - main, 2 - chat, 4 - side, 8 - tutorial
     modalMain = -1;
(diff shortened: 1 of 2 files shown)


ultraviolet-jordan:
Maybe lacking some information, but here says randoms only spawn if not busy:
https://github.com/2004Scape/Server/blob/main/data/src/scripts/login_logout/login.rs2#L26

I guess we don't know if the welcome screen counts or not.

tannerdino:
> Maybe lacking some information, but here says randoms only spawn if not busy: https://github.com/2004Scape/Server/blob/main/data/src/scripts/login_logout/login.rs2#L26
> 
> I guess we don't know if the welcome screen counts or not.

https://youtu.be/T0ulsgorBkY, https://youtu.be/HwGAzcmvF9k both videos seem to spawn the event on different ticks, only assumed this is due to them closing the login interface on different ticks

henke96:
Those videos are from after they got rid of of the LAST_LOGIN_INFO packet though, which causes the client to send close modal: https://github.com/2004Scape/Client/blob/lastmain/client/src/main/java/client.java#L10934

Pazaz claims that they switched to this packet late 2005, which doesn't cause the client to send close modal as far as i can tell: https://github.com/runejs/refactored-client-435/blob/master/src/main/java/org/runejs/client/message/handler/rs435/widget/visibility/ShowFullScreenWidgetMessageHandler.java#L11

I imagine it became a server side interface at that point.

PR #1108, opened 2025-01-18
fix(content): Add missing if_close in luthas dialogue

henke96:
Source: https://www.youtube.com/watch?v=SInMQzx1K1I&t=181s

Pretty sure something is wrong with the engine to require using if_close here, but we already do it in lots of other places.


ultraviolet-jordan:
You would be correct about the issue, I've reported the problem before, I believe it's a problem under `Player#executeScript` function, where it picks up the script after the dialogue suspension, it does not close the dialogue interface. They were too scared to change anything with it, so the running fix is to call `if_close` in those cases.

PR #1111, opened 2025-01-20
fix(content): Dart stats

thesuddensilent:
fix: fixed incorrect stats for darts
range attack bonus
range strength bonus


data/src/scripts/skill_combat/configs/ranged/darts.obj
```diff
@@ -18,7 +18,7 @@ wearpos=righthand
 category=weapon_thrown
 weight=10g
 iop2=Wield
-param=rangebonus,1
+param=rangeattack,3
```
ultraviolet-jordan, line 21: I think you want rangebonus,1

image

data/src/scripts/skill_combat/configs/ranged/darts.obj
```diff
@@ -215,7 +219,7 @@ wearpos=righthand
...
```
ultraviolet-jordan, line 222: Same thing here.

thesuddensilent: both should be fixed now

PR #1113, opened 2025-01-20
fix(content): Handle attempts to use gold/silver bars on anvil

noted-shark:


tannerdino:
Wow nice find, good research as always!

PR #1114, opened 2025-01-20
fix(content): Small drop table fixes

thesuddensilent:
fix: small drop table fix for rune amounts and coins.
Also added a drop table for shadow warrior but it is unfinished.


data/src/scripts/drop tables/scripts/monk_of_zamorak.rs2
```diff
@@ -3,16 +3,18 @@ gosub(npc_death);
...
-if (map_members = true) {
-    def_int $random = random(20);
-
-    if ($random < 1) {
-        obj_add(npc_coord, zamorak_monk_bottom, 1, ^lootdrop_duration);
-    } else if ($random < 2) {
-        obj_add(npc_coord, zamorak_monk_top, 1, ^lootdrop_duration);
-    }
-}
+//obj_add(npc_coord, npc_param(death_drop), 1, ^lootdrop_duration);
```
stomponbabyheads, line 8: Commenting out line 8 makes them not drop bones at all.
Since they don't actually have a drop table it would be better to delete monk_of_zamorak.rs2

thesuddensilent:
monk of zamorak.rs2 file removed permamently because no drops.
stomponbabyheads he advised to do it because it would be better

PR #1121, opened 2025-01-23
fix(content): Rune Mysteries Lost Notes Dialogue

GitDroo:
This changes the script to correctly display Aubury's name and chat-head giving the Notes, instead of the player's.


Pazaz:
ignore CI: builds locally fine. I have to update the lockfile dependency list for the verify jobs to complete. Thanks!

PR #1147, opened 2025-01-30
fix(content): dancing donkey inn bartender dialogue

dhemms:
i used labels from an example from another NPC, where it asks a question and takes you back to previous multiple choice, but with difference choices as you progress. if there is a better way to do this i'd be interested in learning... but it works!

used RSProx to get the correct anims

### DIFF
--- a/data/src/scripts/areas/area_varrock/scripts/bartender.rs2
+++ b/data/src/scripts/areas/area_varrock/scripts/bartender.rs2
@@ -144,7 +144,7 @@ say("Blearrgh!");
 
 [opnpc1,bartender_dancingdonkey] // Dancing Donkey Inn
 ~chatplayer("<p,happy>Hello.");
-~chatnpc("<p,quiz> Good day to you, brave adventurer. Can I get you a refreshing beer?");
+~chatnpc("<p,happy> Good day to you, brave adventurer. Can I get you a refreshing beer?");
 @multi3(
     "Yes please!", bartender_dancingdonkey_yes,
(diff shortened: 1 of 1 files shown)


thesuddensilent:
This NPC does not handle any clue scrolls.
Only Blue Moon in Bartender handles clue scrolls in the Varrock area.

dhemms:
indeedy, that's only for `[opnpc1,bartender_varrock]` and is unchanged by my pull request

multiple bartenders are bundled into bartender.rs2 - I'm unsure why it thinks I'm removing and replacing that line as it was previously the end of the file

data/src/scripts/areas/area_varrock/scripts/bartender.rs2
```diff
@@ -140,4 +140,39 @@ say("Blearrgh!");
...
+~chatplayer("<p,happy>Hello.");
+~chatnpc("<p,quiz> Good day to you, brave adventurer. Can I get you a refreshing beer?");
+@multi3(
+    "Yes please!", bartender_dancingdonkey_yes,
+    "No thanks.", bartender_dancingdonkey_no,
+    "How much?", bartender_dancingdonkey_howmuch
+);
+
+[label,bartender_dancingdonkey_yes]  
+~chatplayer("<p,happty>Yes please!");
```
toomykins, line 155: `happty`

toomykins:
>I'm unsure why it thinks I'm removing and replacing that line as it was previously the end of the file

it only shows as that because of the newline character.

PR #1151, opened 2025-01-30
fix(content): mage arena spells in pvp

dhemms:
this matches player_magic.rs2 and now works as expected


dhemms:
changed to draft, need to fix stat reduction

dhemms:
works as intended with each cape/staff combo

PR #1154, opened 2025-01-31
fix(content): Magic Guild Store delta and stock and Mage Arena delta

puremana:
**What I did**

- Change delta from Magic Guild Store and Mage Arena from 100 -> 20
- Change Nature Rune stock in Magic Guild Store from 50 -> 150
- Change Chaos Rune stock in Magic Guild Store from 50 -> 250
- Change Death Rune stock in Magic Guild Store from 50 -> 150

**Explanation**

When I was playing around with the Magic Guild Store and then looking through Discord, I noticed a screenshot that Tanner had sent through that seems very authentic.

image

I found that the prices in the image and the prices in the shop (it has to be the Magic Guild Store btw, no other shop stocks all those items) were impossible.

This meant that the delta was wrong.

After changing the delta from 100 to 20, which I did because it was the most popular delta for shops, I began playing around and this time found that yes a law rune can be for sale for 59gp (at 26 stock).

Then I found that the nature rune and chaos rune prices were impossible.

I believe this to be because the stock amounts were wrong.

If I increase the stock of natures to 150, chaos runes to 250 (chaos runes at 250 stock is what OSRS stores usually are), and death runes to 150 I find a very interesting result.

I find that at 0 stock, chaos runes are exactly 90gp, and natures at 1 stock are exactly 79gp, and death runes at 1 stock are exactly 119gp.

I saw on the discord that the prices for each of the runes were from the client, so I thought the change had to be in stock and delta.

The change to the mage arena delta was merely because it remained as the only delta in the game still at 100, which I thought is likely wrong and changed it to 20 too. Feel free to change that back if you're not convinced (or I can).

image

image

**Testing**

If you wanna test yourself use `::tele 1,40,48,35,14`

You can search for Tanners original message in the discord by searching `idk if i believe 1k law runes`


thesuddensilent:
wizard tower magic store 2004

we have this image that should show the full stack of this magic guild store tanner found it on some site back in the day

Indio3:
hard to say for sure what full stocks were since images could have been edited, but none of the images have over 50 stock for high level runes, although https://oldschool.runescape.wiki/w/Update:Rune_Shop_Changes does mention rune store stocks were decreased

puremana:
Thanks for the input. That image showing a full stack is interesting, and leaves three scenarios.

1. The picture of prices I'm using is from after 2004 but before the chat updates, in which case the delta is correct.
3. The stock of runes was decreased as Indio3 suggested could be the case in 2005, and the picture you have is from 2005 after the update, in which case my PR is correct.
4. The delta is enormously large to the point that chaos runes at stock 50 can go from 15 -> 90, in which case my entire PR is incorrect.

I'll have a play around with the delta a bit more and see if I can fix it while keeping same stock

thesuddensilent:
Are there any price changes in runes between versions 194-289 depending on the cache?

puremana:
> Are there any price changes in runes between versions 194-289 depending on the cache? --item values--

I don't know how to check that but from my understanding laws are never 59gp or lower again, so prices have to be from this time

Indio3:
> Are there any price changes in runes between versions 194-289 depending on the cache? --item values--

no

thesuddensilent:
ok

Indio3:
If the deltas are 20 for these shops it's probably 20 for the f2p shops as well, the chaos/deaths stocks in those shops will need to be increased so that they're max price at 0 stock (90/180gp), restock times also need be adjusted (all high level runes had their price changed to be 6x more expensive, the restock times were prob 6x longer than current osrs values?)

puremana:
I played around with the delta while keeping stock as it was and couldn't find a value that fits all of the runes in the screenshot.

I came very close a few times. Below is a screenshot using **delta 124** where every value is correct aside from the nature rune which is 79 instead of 78, so I don't think it is possible.

image

Indio3:
I think its likely that they were checked at different times and stock images might not completely match with the checks either

puremana:
> I think its likely that they were checked at different times and stock images might not completely match with the checks either

It's nothing to do with stock images, it's me trying to find any point where the stock _can_ cost that much in the shop with one delta, for all the runes.

I.e if you bump up the delta there would be no stock amount where death runes are exactly 119gp, even though nature runes might be 78gp at 26 stock

Indio3:
> > I think its likely that they were checked at different times and stock images might not completely match with the checks either
> 
> It's nothing to do with stock images, it's me trying to find any point where the stock _can_ cost that much in the shop with one delta, for all the runes.
> 
> I.e if you bump up the delta there would be no stock amount where death runes are exactly 119gp, even though nature runes might be 78gp at 26 stock

If all values are possible with a delta of 20 with the increased stocks I think that's probably correct then, it might be 250 stock for all high level runes (matching OSRS)

thesuddensilent:
Bate varrock1

this picture is really unrelated to the case.
but here is a varrock beta picture probably from december 2003?. considering the spell book is missing god spells completely.
so in chat you can see rune prices with a game message notification.

Indio3:
fairly convinced it is delta 10 with double OSRS stock (500 runes) for all rune shops, would match the elemental prices here perfectly:
https://web.archive.org/web/20050206183221/http://www.runehq.com/cacheguides/viewcityguide00256.htm

thesuddensilent:
kuva
 rune village guide pages have more price infos yanille magic guild
https://web.archive.org/web/20041226180506/http://runevillage.com/guide2Yanille.php

thesuddensilent:
kuva
and tip.it  https://web.archive.org/web/20040806002003/http://www.tip.it/runescape/index.php?page=rs2magic.htm

thesuddensilent:
could it be possible the store should have stacks like this?
air ,water, earth, fire, mind, body =3000 stock
soul = 50 stock
law, nature, chaos, death, blood =1000 stock

Indio3:
for elementals and souls I think those stocks are correct, for the other runes I don't think so cause that'd require a delta of 5 for 0 stock to be 6x price, which wouldn't match any of the images we have

the full stocks for those runes were probably hard to get since i'm guessing the restock time is very slow (probably much slower than what we have it as right now)

puremana:
Oh wow, 5k elemental runes. Interesting.

That screenshot of battlestaffs at 7000-7370 doesn't matter much imo because that is after prices changed and delta definitely could have changed at the same time.

This screenshot you shared is pretty interesting though as it means the stock can't be massive, even with delta 10 as the price would be higher than what is listed

image

thesuddensilent:
that 1000 may very well be correct in our version.?? what you think

since slayer was released in early 2005 runes were needed more and more because of slayer tasks.
and getting them started to be harder and harder.

then a few months later jagex updated the rune restock times to be faster and reduced the stacks??.

Indio3:
if its 1000 the delta is either below 10 (unlikely) or the prices are always maxed below 500 stock (which they are probably always at  ~20 minutes after an update), which I don't think is the case based off the images in the PR post

thesuddensilent:
2005 december
https://web.archive.org/web/20051111053135/http://www.tip.it/runescape/index.php?page=magic_guild.htm

Indio3:
after https://oldschool.runescape.wiki/w/Update:Rune_Shop_Changes I think the delta and stock should match OSRS (seen some 2006 rune shop videos that indicate this)

thesuddensilent:
That's probably the update where the stock was changed to 250.

Indio3:
> This means we have been able to decrease the number of runes in a shop straight after reboot

 yeah probably, had to have been higher before

thesuddensilent:
The magic guild store was probably updated when Slayer was released at the same time Mystic was added.

PR #1157, opened 2025-02-01
fix(content): remove pineapple_plants debug message

dhemms:
Looks like this was used during testing to ensure the plant was being "depleted" whilst picking, but wasn't removed after

screenshot-1738373650


kylmp:
I would just remove the line, it's unnecessary clutter

PR #1172, opened 2025-02-03
Update prospect.rs2

9338088:
sound should be first then 3 second delay til the message


9338088 (commented):
fixed sound was to quick

Indio3:
this will change prospect to a 5t action (p_delay(1) is 2t, p_delay(2) is 3t)

Pazaz:
Fixed in #1288 Ty!

PR #1177, opened 2025-02-04
fix(content): thurgo post-quest dialogue

dhemms:
From RSC Wiki, after the quest:
https://classic.runescape.wiki/w/Transcript:Thurgo#After_The_knight's_sword_quest

_Player: Thanks for your help in getting the sword for me
Thurgo: No worries mate_

From RS3 Wiki, after the quest:
https://runescape.wiki/w/Transcript:Thurgo#After_the_Knight's_Sword

Choice 3: Something else.
_Player: About that sword...
Player: Thanks for all your help getting it for me!
Thurgo: No worries mate._

OSRS Wiki, just for reference:
https://oldschool.runescape.wiki/w/Transcript:The_Knight%27s_Sword
There is no mention of "_No worries mate._" which was used in Classic -> RS3 which suggests the OSRS dialogue was re-written at some point.

I've therefore gone with the RSC Wiki version because there are no other Choices like in RS3, and can presume it would just be: Click Thurgo -> "Thanks for your help in getting the sword for me!" -> "No worries mate." 

RS3 still uses "No worries mate." after completion which suggests this is unchanged from RSC (other than punctuation) and therefore should be the same in 2004.

This may not be 100% authentic, but is a step closer than the pre-quest "Thurgo doesn't appear to be interested in talking." message

### DIFF
--- a/data/src/scripts/areas/area_port_sarim/scripts/thurgo.rs2
+++ b/data/src/scripts/areas/area_port_sarim/scripts/thurgo.rs2
@@ -1,8 +1,12 @@
 [opnpc1,thurgo]
-// todo: Figure out if Thurgo has any possible post-quest dialogue pre-skillcapes
 switch_int (%squire_progress) {
     case 2 : @thurgo_inquire;
     case 3 : @thurgo_special_sword_post_pie;
     case 4, 5, 6 : @thurgo_about_sword;
+    case 7 : @thurgo_squire_complete;
     case default : mes("Thurgo doesn't appear to be interested in talking.");
 }
+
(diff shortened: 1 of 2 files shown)


Indio3:
Since we have the "About that sword..." branch (which classic doesn't have), it might be more correct to send it to that label and check if the quest is complete (or if they have the blurite sword) here: https://github.com/2004Scape/Server/blob/main/data/src/scripts/quests/quest_squire/scripts/quest_squire.rs2#L186

dhemms:
Thanks, will look at putting it under the "About that sword..." branch which seems the cleaner way t

PR #1189, opened 2025-02-04
fix(engine): clear pending action upon failed packet validations

ultraviolet-jordan:
All of these packets call into `player.clearPendingAction();` anyways, but we need to do it on hard failed handled packets, like menu options, visibility, inventory listeners, etc, things that should be impossible for the client to see or do. I did not add these to any of the `delayed` checks.

idk 227


tyler27:
LGTM

tyler27:
yw np anytime

PR #1195, opened 2025-02-05
feat: Save client control settings to localStorage

seditionist:


seditionist:
Had request to add option to show controls again after they were hidden by mistake. Adding it to localStorage as well in that case.

PR #1208, opened 2025-02-06
fix(content): Fix message received when woodcutting without axe

MrAshieB:
Previously, the message was "You do not have an axe which you have the woodcutting level to use.".

This doesn't seem authentic according to images I have seen, including those commented in the code. It does not mention 'woodcutting' in the message authentically.


tannerdino:
Good catch! Completely overlooked that, thank you for the pr

PR #1209, opened 2025-02-06
Fix dragon slayer softlock

MattGould1:
A player (bog) died while fighting elvarg and could not get back to crandor.

bug

I think there were two issues

1. A dialogue issue where Ned asks for a map, despite you handing it to him. This is because we check only for `quest_dragon_ned_given_map` however once sailed to crandor you're set to `quest_dragon_sailed_to_crandor`

2. The dialogue handler for ned would default to `@ned_dragon_slayer_hired` which does not allow the player to re-hire ned. I changed it so that the dialogue handler will first check to see if you have been to crandor. 

Now players should be able to re-hire ned, fix the ship, and meet him in port sarim.

### DIFF
--- a/data/src/scripts/areas/area_draynor/scripts/ned.rs2
+++ b/data/src/scripts/areas/area_draynor/scripts/ned.rs2
@@ -2,16 +2,16 @@
 // Treasure Trail
 if(map_members = true & inv_total(inv, trail_clue_easy_simple_exp021) = 1) {
     @trail_ned;
-} else if (%dragon_ned_hired > 0) {
-    // ned got the job
-    // looks like dragon slayer overrides neds standard dialogue https://youtu.be/kCMB4I_aJJM?t=129
-    @ned_dragon_slayer_hired;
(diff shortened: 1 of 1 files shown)


data/src/scripts/areas/area_draynor/scripts/ned.rs2
```diff
@@ -184,7 +184,7 @@ if ($choice = 1) {
         ~chatnpc("<p,neutral>I said I would and old Ned is a man of his word!"); // osrs
         ~chatnpc("<p,quiz>So where's your ship?");
         @ned_so_wheres_your_ship;
-    } else if (%dragon_progress = ^quest_dragon_ned_given_map) {
+    } else if ((%dragon_progress = ^quest_dragon_ned_given_map | %dragon_progress = ^quest_dragon_sailed_to_crandor)) {
```
kylmp, line 187: no need to change but the extra parenthesis are unnecessary

MattGould1: Oops yeah noted for next time :)

PR #1210, opened 2025-02-06
fix: Shadow Warrior droptable

sentientflare:


thesuddensilent:
https://web.archive.org/web/20041127015309/http://www.tip.it/runescape/index.php?rs2monster_id=115
tip.it november 2004.
marked 45 air, 3 cosmic, 2 death, 2 blood.
unchanged since 2004 even the osrs table matches this.

PR #1241, opened 2025-02-09
fix(engine): lineofsight and lineofwalk commands returning true for members zones in F2P

ultraviolet-jordan:
To fix being able to get out into members areas.
To be clear, the members areas are unloaded and you cannot actually do anything while on a F2P world.

image


tyler27:
LGTM

PR #1246, opened 2025-02-09
fix: Changed fishing synths to be only on start of the action

Devenir377:
Fishing synths should only play once when you start the action. This saw more changes after https://oldschool.runescape.wiki/w/Update:Area_Sounds


tannerdino:
So any proof of this?

Pazaz:
Some discussions ongoing in Discord - a few Oct/Dec 06 proofs. Waiting to hear back from contributors :)

If it's anything like #1247 we can probably remove some of the skill_sound varp uses!

PR #1283, opened 2025-02-12
fix(content): Remove "gay" from chat filter

hikilaka:
Remove "gay" from chat filter:  While aiming for an authentic 2005 experience, censoring "gay" has unintended negative consequences, reinforcing harmful connotations.  This change allows for more inclusive communication.

### DIFF
--- a/data/src/wordenc/badenc.txt
+++ b/data/src/wordenc/badenc.txt
@@ -122,7 +122,6 @@ fuq
 fux
 gangrap
 gangrape
-gay
 gecities
 gecity
 genital


KeepBotting:
Why single out `gay`? If we're going to do this we should prob remove `homo`, `queer`, `lesbian`, arguably `marijuana`, etc

kylmp:
This has always been about preserving the game as it was without exception. Trust me we understand what you are trying to say, but this is just going to stir up drama from both culture war people and from the people dedicated to keeping the authenticity of the recreation.

dhemms:
This was already discussed and although people are understanding, the consensus was to not change the filter:
https://discord.com/channels/953326730632904844/1339183445032701955

It's a slippery slope. It's easier to say "sorry but this is authentic" than have to argue "well if `gay` was removed, why can't these 54 other words be removed too?"

Pazaz:
We want to focus on preserving a pixel-accurate snapshot in time so this is out of scope, and I'm going to close it to prevent any trolls from coming for your thread.

We are solely focused on digital game preservation and this decision does not reflect any personal beliefs from contributors. The game censor was VERY strict back then, causing some funny workarounds to say what you wanted. It also doesn't capture any words added to our vernacular since then.

I appreciate the effort and trying to make a change to better the world around you! I agree that a modern-day live service game should reflect on what language they allow.

PR #1303, opened 2025-02-17
fix(content): skulls no longer last turbo long

tannerdino:


data/src/scripts/skill_prayer/scripts/prayer_interface.rs2
```diff
@@ -123,7 +123,7 @@ return (false);
 %prayer_protectfrommelee = ^false;
 ~headicons_clear;
 if (%pk_skull > 0) {
-    ~pk_skull(%pk_skull);
+    ~pk_skull(sub(%pk_skull, map_clock));
 }
```
kylmp, line 127: Hmm, curious why this is part of `deactivate_prayers`?

kylmp: Maybe it would be better to explicitly clear prayer headicons rather than worry about re-adding headicons that shouldn't have been cleared

tannerdino: Would need to change it every time new headicons are added, easier to just do it this way

kylmp: Needs to be changed anyway whether its new prayer headicons or some other headicon which shouldn't be cleared by draining prayers. Not sure if the former exists though, maybe the red fight pits skull is the only that comes to mind

tannerdino: Actually on second thought you might be right. Relooking at the code, `~deactivate_prayers` would also clear like hint arrows on you and stuff

PR #1315, opened 2025-02-18
chore(content): watchtower completion, debug, typo

dhemms:
Moved `%itwatchtower_progress = ^itwatchtower_complete;` to `itwatchtower_quest_complete`
Standardised quest completion to use `send_quest_complete`
Fixed capitalisation: `it -> It`
Added Watch Tower to `cheat_quest.rs2`, including the separate rewards

It's interesting because, authentically, the rewards are given at the same time the dialogue starts with the wizard. But that means they're separate from the quest completion interface.

### DIFF
--- a/.github/workflows/engine.yml
+++ b/.github/workflows/engine.yml
@@ -1,12 +1,12 @@
 ﻿name: Verify Engine Code
 
 on:
-  pull_request:
-  push:
-    paths:
-    - '.github/workflows/engine.yml'
-    - 'src/**'
-    - 'tools/**'
+  # pull_request:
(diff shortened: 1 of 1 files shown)


Indio3:
Why is the ending order being changed around (varp assignment being moved to queue + inv_add before the delay)? It's intentionally written that way to match the actual game (this change is inauthentic)

dhemms:
Thanks @Indio3 - when updating `cheat_quest.rs2` I realised it wasn't completing the quest which is how I ended up "standardising it" - but good to know the original is authentic!

PR #1316, opened 2025-02-18
fix(content): watchtower typo, added debugging

dhemms:


Indio3:
you can keep the reward being added after the delay (right above where the questpoints was added), it would match the linked video

also you might want to set %itwatchtower_bits = 0 under
```
case 49 : %itwatchtower_progress = 0;
```
as it's also used for part of the quest tracking, so not resetting it wouldn't correctly reset the quest

dhemms:
thanks for the help!

PR #1326, opened 2025-02-18
feat: add login server metrics

hex-agon:


Pazaz:
Does this still work in single-world mode where a management server might already be running on the main thread?
It will work in multi-world setups across diff hosts, I'd guess

hex-agon:
I could start the local server normally but i've added a conditional to only start the managementWeb if the `LOGIN_SERVER` env is set.

PR #1328, opened 2025-02-18
fix(content): uncomment out law and death rune rc multiplier

tannerdino:


Indio3 (commented):
death and law runes don't have a multiplier, would probably be better to just change script to default to 1 if the field count is 0

thesuddensilent:
That 2x law and 2x death are actually an OSRS custom update, they didn't exist in 2013 when the game launched.
surprising it's not listed in the wiki's update list.
it was in a player poll somewhere.

thesuddensilent:
kuva

tannerdino:
I see, thank you both!

PR #1333, opened 2025-02-20
feat(engine): add drop_socket cheat command

locked-chest:
This is useful for testing reconnection logic as there isn't an easy way to ask browser to close a WebSocket.

It silently drops connection so, to the client, it's the same as the network having a problem.

Limited to moderators+.


ultraviolet-jordan:
The client already has `::dropsocket`

ultraviolet-jordan:
The server already has `::serverdrop`

PR #1334, opened 2025-02-20
fix(engine): improve (and simplify) input tracking

locked-chest:
Old code would bail early on the first flush of input events. New code properly waits for all data to be received (up to a limit). It also simplifies quite a lot.

Tested:
- Kicks player due to inactivity exactly same as before
- Processes all input event batches received by the client


ultraviolet-jordan:
It was intended to finish the tracking early so it avoids it to be abusable somehow, we don't want a single player to spend 20ms looping events to submit to a database. I suggested to Pazaz that the events should probably be stored as entire blobs of data instead, and the data be processed later by something else (like a mod tool). But either way, hope this doesn't hurt the server.

ultraviolet-jordan:
You should test what happens and how fast this work is completed on a full world, this is what I do.

PR #1335, opened 2025-02-20
fix(engine): improve input tracking

locked-chest:
- Store raw data from client instead of new row per input event (there could be tens of thousands per tracking event!)
- Introduce rate limit for how much data we're willing to store per tracking session (it's 500 bytes limit per flush, and anecdotal testing suggests we won't get more than a few thousand bytes per tracking session, but if someone is malicious, we cap to 50kb per few minutes).
- Fix bug where we were losing part of the logs due to race between ending the tracking session and pushing data to the DB.

In time when we don't need data stored in the previous table we can easily remove it.

Tested:
- Player is still AFK kicked due to no activity
- Players which are tracked vs untracked do not know the difference (we don't end early or anything)
- We don't parse data if the user isn't being specifically tracked
- Data is stored correctly, and can be replayed correctly.

### DIFF
--- a/prisma/multiworld/migrations/20250220184917_input_report_event_raw/migration.sql
+++ b/prisma/multiworld/migrations/20250220184917_input_report_event_raw/migration.sql
@@ -1,0 +1,9 @@
+-- CreateTable
+CREATE TABLE `input_report_event_raw` (
+    `input_report_id` INTEGER NOT NULL,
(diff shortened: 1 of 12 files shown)


ultraviolet-jordan:
Ye I think doing it as blobs of data is gonna be the "fastest/best" way to go about it on the server. Looping through every individual event for every player just sounds sus.

PR #1353, opened 2025-02-23
feat(engine): add unsafe methods for getting players and npcs in buildarea

MattGould1:


MattGould1:
People noticed that the zombie and shade would have a fire animation and then not appear for 4 ticks. It seems that `npc_delay` prevents the npc from visually spawning?

I'm not sure if this PR is authentic because now the fire animation places and 1 tick later the npc appears and attacks.

ultraviolet-jordan:
The bug is because of a recent change with “entity validation” and inside BuildArea where it searches for npcs in zones, Npc.isValid() returns false when delayed, so then they get skipped when the server is searching for npcs to render for you.

MattGould1:
> The bug is because of a recent change with “entity validation” and inside BuildArea where it searches for npcs in zones, Npc.isValid() returns false when delayed, so then they get skipped when the server is searching for npcs to render for you.

Okay yup, works identically to https://youtu.be/hYvYM82QrN4?si=65jDDwlldAZqY07y&t=222 when commenting out the `this.delayed` check.

Should I go ahead and remove the check? Not sure of the downstream consequences or should we have an entirely new flag for this type of npc_delay that's different to other npc_delay?

ultraviolet-jordan:
There should probably be “unsafe” variants for getNpcs and getPlayers inside the Zone class similar to the other “safe/unsafe” functions. And BuildArea should use the unsafe instead of the safe ones when searching for npcs or players.

MattGould1:
Nice :) locally this behaves identically to the videos I see of randoms spawning. 

1. Animation places
2. NPC is visual but "inactive" not currently targetting you
3. Begins attacking after a few ticks

PR #1358, opened 2025-02-24
fix: Updated Dockerfile - would not find json - /app empty

NoHandlebars87:
-Docker does not bind the volume at build-time, only at runtime, resulting in /app directory being empty

-Replaced VOLUME with COPY to build the Dockerfile

-The volume is still mounted at runtime by docker-compose

Split effort by myself and iPsych0

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,9 +2,8 @@ FROM node:22
 
 RUN apt-get update && apt-get install -y openjdk-17-jre-headless
 
-VOLUME ["/app"]
-
 WORKDIR /app
+COPY . /app
 
 RUN npm install
 


NoHandlebars87:
When running the Dockerfile as is, it would fail with "Could not read package.json: Error: ENOENT: no such file or directory, open '/app/package.json'".

I added COPY ./ /app under WORKDIR, which got the docker running and found the file, however it was still not functioning correctly. 

Psycho suggested commenting out/removing VOLUME [" /app"] while also keeping COPY.

I had commented out VOLUME earlier in troubleshooting, but had had not added COPY simultaneously - once done - this allowed for a successful docker - and a reachable client at host:8888/rs2.cgi

Dockerfile as changed
dockerfile

Successfully running client
docker rs client

Docker running
running docker

PR #1369, opened 2025-02-26
fix(engine): Npc timers, hunts, and interactions/movement have been significantly reworked to better match rs2 engine

mark-b5:
This pull request changes a significant amount of Npc behavior and mechanics to more accurately model the rs2 engine

- Hunts for Npcs have been significantly reworked

  - Hunt delay is processed during the Npc's turn, and is better handled in general
  - findnewmode now runs the `ai_queue` script immediately rather than adding it to the Npc's queue
  - If `findkeephunting` is no, then the Npc will disable its huntmode after consuming a hunt target
  - `Npc.defaultMode()` does a better job of resetting the Npc's relevant hunt variables
  - findnobodynear=no now checks if Npc is rendered by any player before allowing a hunt

- Npc pathing and interactions is a bit different

  - Interaction order has corrected
  - More robust logic for handling Npc walktrigger
  - `Npc.defaultMode()` no longer tries to path the Npc
    - This was causing the Npc to path on the tick they lose target
    - Imps could path two tiles in a single tick!
  - Destination is no longer cleared every tick when the Npc has a targeted mode
  - Npc's max range is no longer checked when it shouldn't be

- Npc timer logic improved
  - Timer is now incremented during an Npc's turn
  - Timer variables are now properly reset when `Npc.defaultMode()` is called
  - Poison logic has been rewritten to be more accurate to rs2

### DIFF
--- a/src/engine/World.ts
+++ b/src/engine/World.ts
@@ -72,6 +72,8 @@ import Environment from '#/util/Environment.js';
 import { printDebug, printError, printInfo } from '#/util/Logger.js';
 import { createWorker } from '#/util/WorkerFactory.js';
 import HuntModeType from '#/engine/entity/hunt/HuntModeType.js';
+import HuntNobodyNear from '#/engine/entity/hunt/HuntNobodyNear.js';
+
 import {
     trackCycleBandwidthInBytes,
     trackCycleBandwidthOutBytes,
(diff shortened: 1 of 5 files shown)


src/engine/entity/Npc.ts
```diff
@@ -64,10 +65,12 @@ export default class Npc extends PathingEntity {
     timerInterval: number = 0;
     timerClock: number = 0;
     regenClock: number = 0;
+    huntClock: number = 0;
     huntMode: number = -1;
-    nextHuntTick: number = -1;
     huntTarget: Entity | null = null;
     huntrange: number = 0;
+    observerCount: number = 0;
```
ultraviolet-jordan, line 72: Why not use the ZoneGrid?

ultraviolet-jordan:
It seems with, using Ranged for example, I can keep pulling npcs away from their spawn farther and farther.
Also we originally made an intential "bug" because of this update, not sure if this was missed in your PR:

https://oldschool.runescape.wiki/w/Update:Lots_more_improvements
*Npcs better at retreating. We've improved the intelligence of the npcs a bit, such that if they are being attacked from beyond the edge of their wander zone, they will run away rather than just standing there stupidly.*

Although that is pretty much up to anyones interpretation...

mark-b5:
> It seems with, using Ranged for example, I can keep pulling npcs away from their spawn farther and farther. Also we originally made an intential "bug" because of this update, not sure if this was missed in your PR:
> 
> https://oldschool.runescape.wiki/w/Update:Lots_more_improvements _Npcs better at retreating. We've improved the intelligence of the npcs a bit, such that if they are being attacked from beyond the edge of their wander zone, they will run away rather than just standing there stupidly._
> 
> Although that is pretty much up to anyones interpretation...

Good catch. I think I need to refactor under what circumstances pathing is being calced when both Npc and Players set their interactions to a target. I'll think it through thoroughly and push a fix for it

mark-b5:
https://github.com/2004Scape/Server/pull/1374

Fixed here

src/engine/entity/Npc.ts
mark-b5: Not accurate to OSRS. OSRS specifically checks if any player has the Npc rendered, i.e. any player within 15 tiles

src/engine/World.ts
```diff
@@ -918,6 +930,10 @@ class World {
                     state.pointerAdd(ScriptPointer.ProtectedActivePlayer);
                     ScriptRunner.execute(state);
 
+                    // Decrement observers of rendered npcs
+                    for (const npc of player.buildArea.npcs) {
```
ultraviolet-jordan, line 934: I think if we have to have this "npc observer count", I would probably put this inside of `removePlayer` instead of here. If a player has a crash in code, then `removePlayer` is always going to be called. This would have to go before the `player.cleanup` is called.

mark-b5: Kk should be good now

PR #1374, opened 2025-02-27
fix(engine): Npcs and Players set and clear their destinations more appropriately

mark-b5:
This pull request stacks on https://github.com/2004Scape/Server/pull/1369 and fixes some further issues in tandem. I reopened the old PR since the issues are now addressed

Before this pull request, every time a Player or Npc interaction is set from runescript, we were doing a path calc to the target. This isn't authentic to osrs and was causing some issues.

- When setting an Npc's interaction, it is unnecessary to immediately set destination, and rs2 skips this. The interaction will be set anyways during the Npc's turn
- When setting a Player's interaction to a PathingEntity, the same thing is true that a path recalc will happen during the Player's turn so it is unnecessary to do upfront. Rs2 similarly skips this
- When setting a Player's interaction to a _NonPathingEntity_ it _is_ necessary to set the Player's destination. Rs2 handles it the following way:
    - In P_OPLOC handler, set Player destination naively to the loc's southwest tile if the player is not already in operable range of the Loc
    - In P_OPOBJ handler, always set the player's destination to the Obj's coordinate
   
Some further things:
- We no longer clear Npc waypoints in NPC_SETMODE handler
- I suspect that p_oplocu exists to make a Player use an item on a Loc. When we implement this, there's some evidence in the game that this should _smart_ path rather than copying the p_oploc implementation of naively pathing
- When multilocs are implemented, we are going to need to refactor the p_oploc check to consider the multiloc's _current type size_ rather than its base size, but `Player.inOperableDistance()` should check the base size. This is confusing but there's plenty of evidence for it


ultraviolet-jordan:
Idk if it's just me but this PR looks exactly like the other PR I think

mark-b5:
It just stacks on the previous PR, but addresses different issues

ultraviolet-jordan:
Which part is the change? It's not really typical to "stack prs" like this so I can't tell where to look.

mark-b5:
The commit that is in this PR and not in the other one

ultraviolet-jordan:
Why is it necessary because when `setInteraction` is called:
```ts
        if (interaction === Interaction.SCRIPT) {
            this.pathToTarget();
        }
```

which `pathToTarget` should be setting the waypoints?

mark-b5:
That line of code doesn't exist anymore. This is explained in the initial 

> Before this pull request, every time a Player or Npc interaction is set from runescript, we were doing a path calc to the target. This isn't authentic to osrs and was causing some issues.

ultraviolet-jordan:
This became different:
idk 233

idk 234

ultraviolet-jordan:
I have a bunch of gifs/videos here for testing purposes:
https://github.com/ultraviolet-jordan/testing-scenarios

ultraviolet-jordan:
Disregard on those gifs, it's because of run enabled/disabled.

ultraviolet-jordan:
I think everything is working right.

mark-b5:
> I think everything is working right.

Thank you for taking the time to test

PR #1377, opened 2025-02-27
fix(content): add period after 'quest' in option select menu when speaking to reldo, and correct dialogue

rust-fox:


rust-fox:
Source: https://web.archive.org/web/20050313124021/http://www.runescapecommunity.com/index.php?showtopic=158384. Credit goes to @Indio3!

PR #1391, opened 2025-02-28
fix(engine): don't convert teleport op to walk/run

Indio3:
Currently, the engine converts a 1/2 tile teleport op to a walk/run, this is incorrect as the client will already convert any <= 2 teleport into a walk anyways if it's not a jump (https://github.com/2004Scape/Client/blob/67f5eda79588149043170088a53465fe3b00aa1a/client/src/main/java/client.java#L10278

This matches OSRS 1, see the p_teleport used here: https://youtu.be/FPNmOhs8jdw?si=K-5L2oh6J3OLuIjs&t=602

Previous:

https://github.com/user-attachments/assets/c786ebb8-b496-488d-8c99-dcb25527ae50

After change:

https://github.com/user-attachments/assets/87dbbfc5-4b19-4103-a26e-1327b07bfdc3

### DIFF
--- a/src/engine/World.ts
+++ b/src/engine/World.ts
@@ -1097,7 +1097,6 @@ class World {
     private processInfo(): void {
         // TODO: benchmark this?
         for (const player of this.players) {
-            player.convertMovementDir();
             player.reorient();
 
             const grid = this.playerGrid;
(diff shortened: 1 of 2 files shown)


tannerdino:
Wonderful, this should fix some inconsistencies like sharks during fishing trawler!

PR #1393, opened 2025-03-01
feat(engine): allow custom debugproc char

MattGould1:
Allows for a custom debugproc char (some people don't have easy access to `~`)

Also removes `NODE_ALLOW_CHEATS` seems to be unused


bundle.js
```diff
@@ -51,17 +50,21 @@ try {
...
-        format: 'esm',
-        write: false,
-        outdir: 'placeholder', // unused but required by esbuild
-        entryPoints: entrypoints,
-        external: externals.concat(esbuildExternals),
-        define: defines,
-        // minify: true,
-        // sourcemap: 'linked',
-    }).catch((e) => { throw new Error(e); });
+    const bundle = await esbuild
```
MattGould1, line 53: My formatter has changed many of the files I touched, could this actually be correct though? I saw the project has a `.prettierrc` and I'm using prettier as my default formatter

Pazaz: Yes, no worries, we have to run prettier over there whole codebase still to normalize it

data/src/scripts/_test/scripts/cheats/cheat_random.rs2
Pazaz, line 1: Did you mean to commit this change to ::~random?

MattGould1: oops :) no let me fix

MattGould1: okay soz, fixed that

src/network/rs225/client/handler/ClientCheatHandler.ts
```diff
@@ -40,13 +40,13 @@ export default class ClientCheatHandler extends MessageHandler<ClientCheat> {
...
         if (player.staffModLevel >= 2) {
-            player.addSessionLog(LoggerEventType.MODERATOR, 'Ran cheat', cheat);    
+            player.addSessionLog(LoggerEventType.MODERATOR, 'Ran cheat', cheat);
         }
 
         if (!Environment.NODE_PRODUCTION && player.staffModLevel >= 3) {
             // developer commands
 
-            if (cmd[0] === '~') {
+            if (cmd[0] === Environment.NODE_DEBUGPROC_CHAR) {
```
kylmp, line 49:
```suggestion
            if (cmd[0] === Environment.NODE_DEBUGPROC_CHAR || '~') {
```
In case the env variable is commented out/undefined have this default fallback

kylmp: nvrm I think its handled in Environment.ts

MattGould1: yeah we use `tryParseString` I guess what could go wrong is that NODE_DEBUGPROC_CHAR is an empty string I'll add this just incase someone does that

PR #1397, opened 2025-03-02
refactor(content): only call ~player_combat_stat on stat change and equipment change

tannerdino:
image

Fairly certain they only need to be calculated on stat change and equipment change. I think there were also some osrs bugs related to this at some point with new weaponry?

Should also help with server tick speed :)

### DIFF
--- a/data/src/scripts/areas/area_combat_training/scripts/combat_training_camp.rs2
+++ b/data/src/scripts/areas/area_combat_training/scripts/combat_training_camp.rs2
@@ -43,7 +43,6 @@ if(%biohazard_progress < ^biohazard_complete) {
 ~open_and_close_double_door2(~check_axis(coord, loc_coord, loc_angle), $side, grate_open);
 
 [oploc1,combat_training_dummy] // Combat training camp dummy
-~player_combat_stat;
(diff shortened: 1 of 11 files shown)


tannerdino:
I noticed something that may be an issue.

https://github.com/2004Scape/Server/blob/8d52194951a96015faea27f1437b02f8c8272e72/data/src/scripts/skill_combat/scripts/player/player_melee.rs2#L29-L36

Maxhit could exponentially increase if you never equip or regenerate a stat. Maybe %com_maxhit is recalculated every hit. Perhaps %com_maxhit doesn't even exist? Is there a source for this?

tannerdino:
> I noticed something that may be an issue.
> 
> https://github.com/2004Scape/Server/blob/8d52194951a96015faea27f1437b02f8c8272e72/data/src/scripts/skill_combat/scripts/player/player_melee.rs2#L29-L36
> 
> Maxhit could exponentially increase if you never equip something or regenerate a stat. Maybe %com_maxhit is recalculated every hit. Perhaps %com_maxhit doesn't even exist? Is there a source for this?

On second thought, perhaps a more simple answer is to just use a local variable or %int

PR #1399, opened 2025-03-02
fix(content): Snare says members on level-up

Gregf36665:


hex-agon:
Snare is members only:
https://oldschool.runescape.wiki/w/Update:PvP_Balancing,_LMS,_and_Elf_Graphics

PR #1402, opened 2025-03-03
fix(content): prevent npcs from tick healing non-recoil damage

### LINKED ISSUE #1403: %com_maxhit should be stored in local var
> I noticed something that may be an issue.
> 
> https://github.com/2004Scape/Server/blob/8d52194951a96015faea27f1437b02f8c8272e72/data/src/scripts/skill_combat/scripts/player/player_melee.rs2#L29-L36
> 
> Maxhit could exponentially increase if you never equip something or regenerate a stat. Maybe %com_maxhit is recalculated every hit. Perhaps %com_maxhit doesn't even exist? Is there a source for this?

On second thought, perhaps a more simple answer is to just use a local variable or %int

_Originally posted by @tannerdino in https://github.com/2004Scape/Server/issues/1397#issuecomment-2693171943_

tannerdino:
- Npc's can no longer tick heal incoming damage, with the exception of ring of recoil damage. However, xp gained and kill credit should still use the capped damage.

From [Video](https://gyazo.com/529a4cbcb261e758c9a4b9764c8a68dc):
image
- Kill credit will now only update on successful hit

- Also fixes #1403

### DIFF
--- a/data/src/scripts/quests/quest_chompybird/scripts/chompy_bird.rs2
+++ b/data/src/scripts/quests/quest_chompybird/scripts/chompy_bird.rs2
@@ -175,7 +175,8 @@ if (%damagestyle = ^style_ranged_rapid) {
 // check hit, give combat xp
 def_int $damage = 0;
 if (~player_npc_hit_roll(%damagetype) = true) {
-    $damage = randominc(min(%com_maxhit, npc_param(max_dealt)));
+    def_int $maxhit = %com_maxhit;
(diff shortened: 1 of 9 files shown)


mark-b5:
Heropoints and xp should still be capped to current hp

tannerdino:
> Heropoints and xp should still be capped to current hp

Good catch

tannerdino:
Notable change I do here is I only call `npc_heropoints` if the hit lands. Is this true in osrs? Right now it looks like you'll be placed into the npc's hero index if you do 0 damage?

PR #1407, opened 2025-03-04
feat(engine): Report review migrations

K-John:
Adds a new `reviewed` column to the `report` table to track whether a report has been reviewed.

## Changes
- Added `reviewed BOOLEAN NOT NULL DEFAULT FALSE` to the `report` model in Prisma schema.
- Generated and included the necessary Prisma migrations.


Jameskmonger:
The [migrations.yml workflow](https://github.com/2004Scape/Server/blob/main/.github/workflows/migrations.yml) should have run on this PR @Pazaz 🤔

PR #1408, opened 2025-03-04
fix(content): A particular dead tree would display "Nothing interesting happens."

tannerdino:
We have no idea how this tree is supposed to work since it has no op3. It was removed in rev 270 (Nov 2004)


ultraviolet-jordan:
If you don't know then why even PR

tannerdino:
> If you don't know then why even PR

I really doubt this had "Nothing interesting happens.". Its not even that hard to code woodcutting with op1 only. Its just different than other trees

ultraviolet-jordan:
kk

PR #1415, opened 2025-03-05
feat(content): witch in draynor manor uses magic now to match old screenshot

tannerdino:
From [archived magazine](https://archive.org/details/PCPowerplay-101-2004-07/page/n13/mode/2up?q=runescape):

image

The witch has the same combat level as dark wizards and the screenshot shows she uses weaken, so I assume she's very similar to dark wizards. 

In rsc and from [runehq](https://web.archive.org/web/20040813232041/http://runehq.com/viewMonster.php?id=34) she drops nothing. 

One potential todo though is that runehq says she can spawn in the middle of the wilderness 👀

### DIFF
--- a/data/src/scripts/_unpack/all.npc
+++ b/data/src/scripts/_unpack/all.npc
@@ -4434,7 +4434,12 @@ model4=model_353_idk
 model5=model_428_idk
 model6=model_361_obj_wear
 model7=model_377_obj_wear
-// TODO respawnrate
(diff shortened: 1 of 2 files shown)


thesuddensilent:
leave this information to default for both witch npc.
wanderrangen=5
maxrangen=7

tannerdino:
> leave this information to default for both witch npc. wanderrangen=5 maxrangen=7 and tip.it mentions that witches are not aggressive. (2005)

Witch won't aggro here, no huntmode set

PR #1417, opened 2025-03-05
refactor(cheat_quest.rs2): modularize quest debug commands

rust-fox:


rust-fox:
@tannerdino @NormoZenyte I've refactored the quest debug commands with several improvements:

* Restructured the command menu with four options: "Complete All Quests", "Reset All Quests", "Select Individual Quest" (preserves original functionality), and "Cancel"
* Added convenient shortcuts: ::~cq for "Complete All Quests" and ::~rq for "Reset All Quests"
* Fixed quest completion order:
  - Quests now complete sequentially with proper NPC interactions
  - Moved NPC-dependent quests (like Druidic Ritual with Kaqemeex) to resolve dialogue interruption issues
  - Temple of Ikov and Tourist Trap now complete properly at the end of the sequence
* Fixed small issues:
  - Restored original quest enum order with "Black Knight's Fortress" at val=0 and "Cook's Assistant" at val=1
  - Added missing "%ikov_progress = 0;" reset when using "Select Individual Quest"
  - Fixed "%ikov_icearrowchest_coord = null;" which was incorrectly overwritten with "0_0_0_0_0" (conflicting with commit b4a01c0)
* Improved menu text for better clarity

The code is now more maintainable and provides a better user experience.

Any thoughts or feedback would be highly appreciated. :)

PR #1418, opened 2025-03-05
fix(content): An extra space in Lancelot dialogue

EricBartusch:


ultraviolet-jordan:
Was there any discussion about this one or blindly merged?

PR #1420, opened 2025-03-05
fix(logging): log milestone total levels even if gaining multiple lvls

kylmp:
Will make the session log about reaching 250 total, even if you were to go from 249 to 251 total in one tick


ultraviolet-jordan:
Add a `break;` in the loop to end the loop early.

PR #1422, opened 2025-03-05
fix(content): Another.. extra space in Lancelot dialogue

AmVoidGuy:
The one case was fixed but the same line in the other case wasn't.


tannerdino:
I agree with Jordan, we need to double check these with the relevant sources. It's happened quite a bit where the original dialogue had double spaces at the beginning of certain sentences.

AmVoidGuy:
Earliest source I see is this [video](https://youtu.be/HW1GxGGWD-0?si=b5rcsTFmVu8YZu6V&t=109 ) from April 2008.  


Put a screenshot from how it currently is to show the difference. 
screenshot-Lancelot

ultraviolet-jordan:
Yeah and looks like according to RSC, it is single spaced.

PR #1424, opened 2025-03-05
fix(content): Genie and mysterious old man were missing special dialogue

tannerdino:
Since mysterious old man has the same dialogue here as genie, its safe to assume that they have the same 'no target' dialogue.

- Talking to a mysterious old man that isnt yours, from [video](https://youtu.be/U_TRFGTfIY0?t=276):

VS--YouTube-RunescapefishingTrippoonage-4’36”

- Talking to a genie that isnt yours *and* talking to a genie that cant find it's target, from [video](https://youtu.be/qNQYgiNMBes?t=134):

VS--YouTube-RunescapeGenieGlitch-2’14”
VS--YouTube-RunescapeGenieGlitch-2’00”


mark-b5:
good commit

Issue #800, opened 2024-08-25
Big Chompy Bird Hunting cooking too early in quest


Jameskmonger:
There are many different dialogues on OSRS if you try to cook (anything) on the spit-roast too early in the quest. I have an RSProx capture for this.

PR #1427, opened 2025-03-06
fix(content): Various quest fixes

Indio3:
- fixed dialogue and adding missing parts of chompy bird quest (buying back bow, releasing toads, cooking chompy after quest), fixes https://github.com/2004Scape/Server/issues/800
- Change picking up black cog to require use for bucket, don't automatically pick up and set a varp instead (https://youtu.be/0hBctCQKW88?si=A_Px4u6qyCfYh64Y&t=89)
- Change a few quest npcs to use last attacker for progression rather than heropoints
- fix a tutorial mesbox to match old magazine image
- add cannon placement check to fight arena


tannerdino:
Did you end up testing some stuff with chompy combat? I dont have a good account to test it. Wondering if it should check autocast, and 2-3 mes (from ~player_in_combat_check). I refactored it a bit to match up with combat a bit but I saw:

`// no aprange stuff at all, cb style doesn't matter`

Comment, which makes me want to test it fully

Indio3:
> Did you end up testing some stuff with chompy combat? I dont have a good account to test it. Wondering if it should check autocast, and 2-3 mes (from ~player_in_combat_check). I refactored it a bit to match up with combat a bit but I saw:
> 
> `// no aprange stuff at all, cb style doesn't matter`
> 
> Comment, which makes me want to test it fully

Didn't test it fully yet, I can check autocast but I don't think it has the duplicate messages from player_in_combat check

thesuddensilent:
Uploading java_ELAbO7WlAZ.mp4…

thesuddensilent:
https://github.com/user-attachments/assets/a5f6c29b-1109-4101-af36-abe7b52f71e5

tannerdino:
Does something like this look about right? I tested it on local and it seems to match what im seeing. With this code all the checks before p_aprange will run from 10 tiles away, is this accurate? We might need to check all of the checks to see if they respect the weapons attackrange

```java
[apnpc5,chompy_bird] @player_combat_chompybird_start;
[opnpc5,chompy_bird] @player_combat_chompybird_start;

[label,player_combat_chompybird_start]
if (~player_in_combat_check = false) {
    return;
}
def_obj $rhand = inv_getobj(worn, ^wearpos_rhand);
if ($rhand = null) {
    // todo confirm for 2004
    mes("You'll need a weapon to try and attack this beast.");
    return;
}
if (%damagetype ! ^ranged_style) {
    // todo confirm for 2004
    mes("The Chompy Bird is too quick for your melee weapon.");
    return;
}
if ($rhand ! ogre_bow) {
    // todo confirm for 2004
    mes("Your ranged weapon isn't powerful enough to hurt the Chompy bird.");
    return;
}
def_obj $ammo = inv_getobj(worn, ^wearpos_quiver);
if ($ammo = null) {
    // https://youtu.be/Qc-4z_z2JLk?si=Bh2bAhDlFQxedAqg&t=446
    mes("There is no ammo left in your quiver");
    return;
}
if ($ammo ! ogre_arrow) {
    // todo confirm for 2004
    mes("You can't use that ammo with your bow.");
    return;
}
def_int $attackrange = ~player_attackrange($rhand);
if (npc_range(coord) > $attackrange) {
    p_aprange($attackrange);
    return;
}
@player_combat_chompybird($rhand, $ammo);
```

Indio3:
Yeah that works, i'll test with https://oldschool.runescape.wiki/w/Comp_ogre_bow

thesuddensilent:
composite bow can only shoot ogre arrows and brutal arrows. but with just ogre bow you can only shoot ogre arrow.
if i understand correctly.

i always use brutal arrows because they are much better than ogre arrows.

Indio3:
```
if ($ammo = null) {
    // https://youtu.be/Qc-4z_z2JLk?si=Bh2bAhDlFQxedAqg&t=446
    mes("There is no ammo left in your quiver");
    return;
}
if ($ammo ! ogre_arrow) {
    // todo confirm for 2004
    mes("You can't use that ammo with your bow.");
    return;
}
```
these checks onward will respect the attackrange, the ones before that are all default (10 tile)

tannerdino:
Sounds good then good testing!

PR #1429, opened 2025-03-07
fix(engine): Locs and Objs are now tracked in a way that better mirrors RuneScape behavior

mark-b5:
Zones used to track Locs and Objs using a ZoneEntityList which was a custom data structure that made it impossible to mirror authentic RuneScape behavior. To be specific, the order that Locs and Objs are added to the Zone has to be preserved and readily available, and ZoneEntityList did not accomplish this.

Zones now track Locs and Objs using our LinkedList implementation. This should prove to be the best way to achieve authentic RuneScape behavior.

`Entity` is now a child of `Linkable`. Players and Npcs are not yet stored to a Zone via a LinkedList, but we want to move in that direction. For now, they will have unused Linkable capability

This is largely a structural change to achieve better behavior _with future updates,_ but one immediate behavioral change is that **Zones will now correctly delete Objs when the Obj limit of 129 per Zone is reached.** Before this change, we were sorting and deleting the lowest alch-value Obj in the Zone when the limit was reached. This isn't authentic behavior. **Zones will now delete the oldest dynamic Obj in the Zone when the limit is reached**, which is authentic behavior

Don't merge this change into `Live` immediately. Let's get ample testing on it in the staging world


mark-b5:
Fixed merge + new lint rules

ultraviolet-jordan:
go to the trawler and do rewards for like 2000 players and huge items drop all over the floor.

ultraviolet-jordan:
can you share any evidence that you have that objs are not sorted by lowest alch-value?

mark-b5:
Firm evidence is probably lost to history, but it's a relatively well known thing that they changed it to lowest alch-value sometimes in 2015 osrs to combat people dropping cheap items to delete people's untradeables like fire capes in pvp

tannerdino:
> Firm evidence is probably lost to history, but it's a relatively well known thing that they changed it to lowest alch-value sometimes in 2015 osrs to combat people dropping cheap items to delete people's untradeables like fire capes in pvp
> 
> For things like obj_findall (not yet implemented in 04) in osrs we also still need to preserve addition order to match behavior

People were doing this to UIMs too, when they were death piling. I wonder if they ever mentioned this change in a blog post

mark-b5:
When calling `loc_change` in runescript, it now changes the type of the Loc in memory, rather than constructing a new one. This is testably authentic to how osrs handles it

- Locs have a new method `change`
    - When called on a dynamic loc, this will overwrite the `info`  bits
    - When called on a static loc, this will set the `tempinfo` bits
- Player now validates against original loc `type` before allowing interactions
- World and Zone have functionality to keep track of changed static locs

mark-b5:
`loc_add` now runs through `World.changeloc()` if there is a layer conflict. This helps us replicate a bunch of OSRS behavior!!!

Further, I sped up all Loc and Obj events by 1 tick. All of our respawn/despawn/reveal events were a whole tick slow in 2004scape compared to OSRS, but they now match. This WILL speed up tree and rock respawns by 1 tick compared to before, but this is probably an authentic change

Anyway here's some new behavior that now matches osrs engine :)

https://github.com/user-attachments/assets/99be1f50-987e-4f2f-ad28-cf5a05e65585

https://github.com/user-attachments/assets/5bd97567-f5d2-4744-b359-5a7bed6a179d

mark-b5:
cheat_zone.rs2 has the following debugprocs to test a few features:

`::~drop_items`, `::~add_locs`, `::~loc_add_del`, `::~loc_add_static`, `::~loc_add_dynamic`,  `::~loc_change_dynamic`,

PR #1438, opened 2025-03-08
feat(content): add huntmode to thieves in monkfriend quest

MattGould1:


data/src/scripts/_unpack/all.npc
```diff
@@ -5302,6 +5302,8 @@ param=attack_anim,human_sword_stab
 param=defend_anim,human_sword_defend2
 param=attack_sound,stabsword_stab
 param=retreat,8
+huntmode=cowardly
```
MattGould1, line 5305: https://youtu.be/rsKPx_gpXSs?si=_kpWqA-k6csJLOge&t=168

data/src/scripts/_test/scripts/cheats/cheat_monks_friend.rs2
```diff
@@ -0,0 +1,9 @@
+[debugproc,setup_monks_friend]
```
MattGould1, line 1: are these welcome or is it just bloat?

data/src/scripts/_unpack/all.npc
```diff
@@ -5302,6 +5302,8 @@ param=attack_anim,human_sword_stab
...
+huntrange=2
```
MattGould1, line 5306: I wasn't really sure about hunt ranges, is there any info on general guidelines for this?

ultraviolet-jordan: The only way to know is to login OSRS and test it out for yourself. There are no known leaks or dumps or collections for any type of npc ranges, well I guess we are the collection now...

data/src/scripts/_test/scripts/cheats/cheat_monks_friend.rs2
ultraviolet-jordan: debugprocs are welcome if they make sense in some way to have, either way nobody really cares.

kylmp: These "setup account to start a quest" debug procs could be useful, but should be more standardized maybe all in one place, or even added to the `::~quest` cheat as a 3rd option (complete, start, reset)

Could be a task for someone to tackle later though

data/src/scripts/_unpack/all.npc
Matt-Gould29: It's an educated guess based on https://youtu.be/rsKPx_gpXSs?si=_kpWqA-k6csJLOge&t=168

In the video, you see one thief follow until the door.

I decided to reduce the hunt range to 1. There are different thieves and now the 2 lv 14s will follow like the screenshots below

Screenshot 2568-03-08 at 13 46 26

Screenshot 2568-03-08 at 13 54 37

data/src/scripts/_test/scripts/cheats/cheat_monks_friend.rs2
Matt-Gould29: okay cool, I'll keep this in mind as I play test other quests

rust-fox: > These "setup account to start a quest" debug procs could be useful, but should be more standardized maybe all in one place, or even added to the `::~quest` cheat as a 3rd option (complete, start, reset)
> 
> Could be a task for someone to tackle later though

If we'd like to continue what was added to [d5cd8eb](https://github.com/2004Scape/Server/pull/1417/commits/d5cd8eb495dfd64d3d10b1303fd9dab7dbe4268e), I believe the quest.enum file would need be updated, and then we can add it to the queue.

PR #1454, opened 2025-03-12
fix(content): Shop changes

thesuddensilent:
The shop stacks of the shops changed here now match the early 2006.
but I don't know if the stacks were the same in 2004 or 2005.
because the shops stacks are significantly different compared to the runescape classic.

### DIFF
--- a/data/src/scripts/areas/area_ardougne_east/configs/ardougne_east.inv
+++ b/data/src/scripts/areas/area_ardougne_east/configs/ardougne_east.inv
@@ -86,6 +86,7 @@ stock5=mithril_platebody,1,3000
 
 [generalshop12]
 // empty vials could be a default stock of 300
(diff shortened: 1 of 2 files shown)


tannerdino:
You're probably right about jatix's herblore shop. Screenshot from 2004:
image

Heres another 2004 screenshot of the ardougne general store:
image

We can probably match up the price of the vial with the shop restock?

Heres also this feb 2006 source:
image

thesuddensilent:
do you notice on that guide page it says vial of water and we have a picture where there are empty vials.
and osrs also has empty ones. so did jagex change empty vials to vials of water at some point? but when was this change made?

thesuddensilent:
do you notice on that guide page it says vial of water and we have a picture where there are empty vials.
and osrs also has vial of waters. so did jagex change empty vials to vials of water at some point? but when was this change made?
(ardougne shop)

PR #1455, opened 2025-03-12
fix(poison): Reworked player poison implementation

pelippper:
Reworked poison to work more authentically. Instead of there being 2 variables, %poison and %antipoison, there is just one variable, %poison. This variable is negative if the player has poison immunity, and positive if the player is poisoned. In addition, added poison immunity to strange fruit to work as described by Ash https://archive.is/6SmRo.

### DIFF
--- a/data/src/scripts/skill_combat/scripts/poison.rs2
+++ b/data/src/scripts/skill_combat/scripts/poison.rs2
@@ -64,9 +64,7 @@ if (stat(hitpoints) = 0) {
 cleartimer(poison);
 
 [proc,set_poison_login]
-if (%poison > 0) {
-    ~set_poison_timer;
-}
+~set_poison_timer;
 
 [proc,set_poison_timer]
 if(gettimer(poison) < 1){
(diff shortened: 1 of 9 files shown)


data/src/scripts/skill_combat/scripts/poison.rs2
```diff
@@ -3,19 +3,25 @@ if (map_members = false) {
...
     return;
 }
-if (%antipoison < map_clock) { 
+
+if (%poison > -1) { 
     if (%poison = 0) {
         mes("You have been poisoned!");
     }
     settimer(poison, 30);
+    %poison = $severity;
```
Lucwousin, line 12: I think this should be %poison = max(%poison, $severity)

pelippper: Updated, thanks!

pelippper:
Also updated each call to settimer(poison,30) to first check that the poison timer isn't already active

pelippper:
Fixed inequality mistake, removed antipoison logout/login procs, added proc for setting poison timer

pelippper:
fixed a few syntax errors I made, oopsie

mark-b5:
Needs the following
- Call ~set_poison_timer proc when setting immunity
- Don't call ~set_poison_timerr proc from inside poison timer
- Change poison_damage to be a queue rather than a proc

pelippper:
Fixed the 3 issues raised by b5 above, tested in-game, working as intended

mark-b5:
Looks good to me

PR #1459, opened 2025-03-12
feat(engine): free and multi zones data storage improvements

ultraviolet-jordan:
Instead of storing per tile in the sets, store it per zone index.
This decreases the amount stored within them (it's basically `original length/64`).
```
423616 (free) --> 6619 (free)
186816 (multi) --> 2919 (multi)
```
This improves the server startup time about 2 seconds on my pc (1 second startup time).
According to console, this saves about 20mb of memory but idk...
I kept all comments in the csv file in the proper spots.

one
two
three


mark-b5:
Looks good. I think they converted multicb zones to use areas after 06 which would have then allowed them to go tile by tile

ultraviolet-jordan:
> Looks good. I think they converted multicb zones to use areas after 06 which would have then allowed them to go tile by tile

I think they're supposed to be individually marked in jm2 files with the other map info, but we can worry about that later i think

thesuddensilent:
That area marked in yellow probably doesn't belong to our version's multicombat area because as shown in the picture, the top 8x8 square doesn't have multicombat.
and I don't see any significant reason why it would in the lower squares either.
the picture is from the 2005 Easter event.

Probably added when jagex updated the region area 47_51 in connection with farming and the graphics updated the area at the same time.
because the multicombat symbol appears in the materials after the graphics update.


kuva
easter event 2005 1

thesuddensilent:
kuva

ultraviolet-jordan:
I have removed those 3 for now then unless someone says otherwise:
map-2560-2688

thesuddensilent:
ok

PR #1472, opened 2025-03-16
fix(engine): Sped up Loc and Obj events by 1 tick to match OSRS, this will make resources respawn slightly faster

mark-b5:
Our `loc_add`, `loc_change`, `loc_del`, `obj_add`, and `obj_del` commands all processed 1 tick slow compared to OSRS. This made things that took, for example, 500 ticks end up being 1 tick too slow. Also, our resource respawns were likely 1 tick too slow

However, we had a lot of content that was written to match OSRS assuming a 1 tick slow event. Since we sped up every event by 1 tick, we have to manually slow down over 150 examples in content where the behavior matched previously 😬

The majority of this PR is sifting through 90+ files and checking whether or not we need to slow down these commands by 1 tick. For every case of `loc_add`, `loc_change`, and `loc_del` I have added a comment that explains whether I chose to slow it down to match previous behavior, or leave it alone to let it be sped up 1 tick by the engine change

@Indio3 @tannerdino Please scroll through files changed if you can

### DIFF
--- a/data/src/scripts/general_use/scripts/coffins.rs2
+++ b/data/src/scripts/general_use/scripts/coffins.rs2
@@ -1,9 +1,11 @@
 [oploc1,loc_398]
 mes("The coffin creaks open...");
 sound_synth(coffin_open, 0, 0);
+// Temp note: dur does not need updated ? @indio
 loc_change(loc_399, 485);
 
 [oploc1,loc_399]
 mes("You close the coffin.");
 sound_synth(coffin_close, 0, 0);
+// Temp note: dur does not need updated ? @indio
 loc_change(loc_398, 485);
--- a/data/src/scripts/general_use/scripts/coffins.rs2
+++ b/data/src/scripts/general_use/scripts/coffins.rs2
@@ -2,10 +2,10 @@
 mes("The coffin creaks open...");
 sound_synth(coffin_open, 0, 0);
 // Temp note: dur does not need updated ? @indio
-loc_change(loc_399, 485);
+loc_change(loc_399, 500);
 
 [oploc1,loc_399]
 mes("You close the coffin.");
 sound_synth(coffin_close, 0, 0);
 // Temp note: dur does not need updated ? @indio
-loc_change(loc_398, 485);
+loc_change(loc_398, 500);
--- a/data/src/scripts/general_use/scripts/trapdoors.rs2
+++ b/data/src/scripts/general_use/scripts/trapdoors.rs2
@@ -2,7 +2,7 @@
 mes("The trapdoor opens...");
 sound_synth(door_open, 0, 0);
 // Temp note: dur does not need updated ? @indio
-loc_change(loc_1570, 485);
+loc_change(loc_1570, 500);
 
 [oploc1,loc_1570]
 mes("You climb down through the trapdoor...");
@@ -14,4 +14,4 @@ p_telejump(movecoord(coord(), 0, 0, 6400));
 mes("You close the trapdoor.");
 sound_synth(door_close, 0,0 );
 // Temp note: dur does not need updated ? @indio
-loc_change(loc_1568, 485);
+loc_change(loc_1568, 500);
--- a/data/src/scripts/general_use/scripts/trapdoors.rs2
+++ b/data/src/scripts/general_use/scripts/trapdoors.rs2
@@ -1,6 +1,7 @@
 [oploc1,loc_1568]
 mes("The trapdoor opens...");
 sound_synth(door_open, 0, 0);
+// Temp note: dur does not need updated ? @indio
 loc_change(loc_1570, 485);
 
 [oploc1,loc_1570]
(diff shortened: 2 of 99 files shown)


data/src/scripts/quests/quest_ikov/scripts/ikov_dungeon.rs2
```diff
@@ -96,6 +97,7 @@ if(last_useitem = ikov_lever) {
 p_arrivedelay;
 anim(human_leverdown, 0); 
 sound_synth(lever, 0, 0);
+// Temp note: dur does not need updated ? @indio
```
Indio3, line 100: yes (19t)

data/src/scripts/quests/quest_ikov/scripts/ikov_dungeon.rs2
```diff
@@ -133,7 +135,8 @@ if(%ikov_progress = ^ikov_disabled_trap) %ikov_progress = ^ikov_pulled_lever;
 p_arrivedelay;
 anim(human_leverdown, 0);
 p_delay(0);
-loc_change(loc_88, 5);
+// Temp note: dur updated ? does it need ??
```
Indio3, line 138: yes this did need update, this is 5t on osrs

data/src/scripts/quests/quest_upass/scripts/kardia.rs2
```diff
@@ -86,6 +86,7 @@ p_arrivedelay;
 anim(human_openchest, 0);
 sound_synth(chest_open, 0, 0);
 p_delay(0);
+// Temp note: dur does not need updated ?? @indio I assume
```
Indio3, line 89: Correct (same should apply to the other 20t chests)
```
[2468]
    [message_game] type=mesbox, message="You search the chest and find a doll, a book and some potions. However, you don't have enough room to take them."
    [loc_add_change] id=3368, coord=(2157, 4564, 1), shape=10, rotation=0

[2487]
    [loc_add_change] id=3272, coord=(2157, 4564, 1), shape=10, rotation=0
```

data/src/scripts/minigames/game_partyroom/scripts/partyroom.rs2
```diff
@@ -132,7 +133,8 @@ p_walk(loc_coord);
 p_delay(0);
 anim(seq_794, 0);
 p_delay(0);
-loc_change(loc_param(next_loc_stage), 5);
+// Temp note: dur updated, does it need ? @indio
+loc_change(loc_param(next_loc_stage), 6);
```
Indio3, line 137: idt it matters cause it gets deleted by end of script

data/src/scripts/quests/quest_grandtree/scripts/king_narnode.rs2
```diff
@@ -215,6 +215,7 @@ switch_int(%grandtree_progress) {
                 }
                 npc_facesquare(movecoord(^narnode_trapdoor_coord, -1, 0, -1));
                 if(loc_find(movecoord(^narnode_trapdoor_coord, -1, 0, -1), loc_2446) = true) {
+                    // Temp note: dur does not need updated ? @indio
```
Indio3, line 218: correct, this is 9t on osrs

data/src/scripts/general_use/scripts/trapdoors.rs2
```diff
@@ -1,6 +1,7 @@
 [oploc1,loc_1568]
 mes("The trapdoor opens...");
 sound_synth(door_open, 0, 0);
+// Temp note: dur does not need updated ? @indio
```
Indio3, line 4: idk where 485 comes from, but tested these on osrs and it should be 500 (499t)

data/src/scripts/general_use/scripts/coffins.rs2
```diff
@@ -1,9 +1,11 @@
 [oploc1,loc_398]
 mes("The coffin creaks open...");
 sound_synth(coffin_open, 0, 0);
+// Temp note: dur does not need updated ? @indio
```
Indio3, line 4: same as trapdoors, tested on osrs and got 500 (499t)

data/src/scripts/general_use/scripts/trapdoors.rs2
mark-b5: Fixed, thanks

data/src/scripts/general_use/scripts/coffins.rs2
mark-b5: Fixed, thanks

mark-b5:
Only thing we need then is @tannerdino to glance over resource respawn rates. They're all going to be 1 tick faster than before

tannerdino:
> Only thing we need then is @tannerdino to glance over resource respawn rates. They're all going to be 1 tick faster than before

Will do shortly!

data/src/scripts/skill_woodcutting/scripts/old_woodcut.rs2
tannerdino, line 1: In osrs level 1 trees last randomly between 59 to 98 ticks. Which is like `add(60, random(40))`, this wouldve been after the resource respawn rate change... So maybe it was literally just `add(120, random(80))`, 119 to 198 ticks (doesnt match our code)

PR #1477, opened 2025-03-16
refactor: extract processCsv to reduce duplicate code

brandly:
I don't have all of the context here, but I saw this csv processing going on and there's a bunch of duplicate code. I extracted the duplicate code into a function called `processCsv`.

I can improve these names if you think there's something better!


Pazaz:
Thanks for the cleanup!

PR #1478, opened 2025-03-16
fix(engine): Truncate stat_subs percentage calculation

xylphie:
Implementation for stat drains seems to be incorrect.

Zammy flames drains one more point than expected

osrs: https://imgur.com/a/ci5dbP0
pre-fix: https://imgur.com/a/zFv737U

npc reduction: https://imgur.com/a/QbuS8x4


thesuddensilent:
I didn't even know that stat changes are visible through monster examination.

mark-b5:
We should probably get rid of the % parameter in the stat_sub commands altogether and just do the math in runescript and then pass a constant

PR #1479, opened 2025-03-16
fix(engine): Truncate statsub percentage calculations

xylphie:
Implementation for stat drains seems to be incorrect.

Zammy flames drains one more point than expected

osrs: https://imgur.com/a/ci5dbP0
pre-fix: https://imgur.com/a/zFv737U

npc reduction: https://imgur.com/a/QbuS8x4

### DIFF
--- a/src/engine/script/handlers/NpcOps.ts
+++ b/src/engine/script/handlers/NpcOps.ts
@@ -436,7 +436,7 @@ const NpcOps: CommandHandlers = {
 
         const npc = state.activeNpc;
         const current = npc.levels[stat];
-        const subbed = current - Math.trunc(constant + (current * percent) / 100);
+        const subbed = current - ((constant + (current * percent) / 100) | 0);
         npc.levels[stat] = Math.max(subbed, 0);
     }),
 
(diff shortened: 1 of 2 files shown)


ultraviolet-jordan:
Try this:

`const subbed = current - ((constant + (current * percent) / 100) | 0);`

mark-b5:
image


From this image we can infer that the `percent` should calc off base stat rather than current stat. `stat_heal(hitpoints,0,100)` should heal you to full hp, but right now it only doubles your current hitpoints.

`stat_sub` implementation should match this. `stat_sub(hitpoints,0,10)` should drain you 10% if your base hp, not your current

ultraviolet-jordan:
The `stat_add` and npc_statadd` commands probably also need the same fix bcuz they also do the `/ 100` like these do. 
If these need to be reworked then can probs do that in another PR.

mark-b5:
> The `stat_add` and `npc_statadd` commands probably also need the same fix bcuz they also do the `/ 100` like these do. If these need to be reworked then can probs do that in another PR.

I guess, but idk if I see the reason to merge this one if the implementation still isn't correct

ultraviolet-jordan:
tenor

mark-b5:
Whatever, no reason not to merge it. But if she has more time there's more to do here if she wants :)

xylphie:
I'll fix the other two, one sec

xylphie:
I updated the other 4 :) All other percentage/division calcs in ts seem to have | or equivalent functionality. 
`const weightKg = this.runweight / 1000;` doesn't, but that could be tested at another time if it hasn't been already

mark-b5:
We can merge this and then I'll write a GH issue for what needs to happen further

PR #1484, opened 2025-03-17
refactor(engine): (npc)statadd/heal/sub now uses base level for percentage calcs

xylphie:
fixed wines of zammy too. drains 1 + 5% of current level.

`stat_sub($stat, calc(abs($constant) + ((stat($stat) * abs($percent)) / 100)), 0);` for consume_effects because consumed stat reductions use current level instead of base


ultraviolet-jordan:
In `STAT_ADD` and `STAT_HEAL` it would be cleaner if they did `PlayerStat.HITPOINTS` instead of `3`.

Also found a bug in `NPC_STATHEAL` and `NPC_STATADD` are checking `stat === 0` which used to be hitpoints but is now attack. So these can be the same using `NpcStat.HITPOINTS` instead of `0` and `stat`.

`NPC_STATHEAL` does a `===` while all the others do a `>=` check, so I would change that to be `>=` like the others.

```ts
        if (stat === PlayerStat.HITPOINTS && player.levels[PlayerStat.HITPOINTS] >= player.baseLevels[PlayerStat.HITPOINTS]) {
            player.heroPoints.clear();
        }
```

```ts
        if (stat === NpcStat.HITPOINTS && npc.levels[NpcStat.HITPOINTS] >= npc.baseLevels[NpcStat.HITPOINTS]) {
            npc.heroPoints.clear();
        }
```

PR #1488, opened 2025-03-17
feat(engine): stat_boost and stat_drain commands

tannerdino:
stat_boost is confirmed, (guthix rest uses it). Stat_drain is guessed, it functions the same as stat_sub except it uses current level of the stat instead of base level of the stat


data/src/scripts/quests/quest_zombiequeen/scripts/quest_zombiequeen.rs2
```diff
@@ -1204,6 +1204,7 @@ if ($op = 2) {
 
 [label,zqcrystal_activate]
 def_coord $central_coord = 0_45_48_36_20;
+// todo: confirm if this is subtracts from stat base or stat @indio
```
tannerdino, line 1207: @Indio3 I dont know if you remember this

Indio3: stat, it stopped lowering once you were <= 24 iirc

tannerdino: Alright ill change it to stat_drain then

PR #1491, opened 2025-03-17
feat(content): More Duel Arena messages

manpaint:
Replaced some of Tanner's message guest with RS3 messages


tannerdino:
Good work! It hurt me having to guess these, cool to see how close I was. Some of these I just assumed existed :D

PR #1494, opened 2025-03-18
fix(content): Cannon fixes

tannerdino:
Fixes and additions:
- 1/4 chance to destroy cannon at kbd instead of 1/10
- Match cannon decay behavior with osrs
- Give 2 ranged xp and no hp xp per damage for cannons
- Special cannon mes's added
- Special nulodion dialogue if partial cannon is setup
- Check line of walk before placing cannon
- loc_add for cannon parts changed to loc_change

Some cannon todos:
- Test on staging world if you can steal other peoples cannons with world hopping (authentic: https://youtu.be/yTiq3TRZrys)
- Retest cannon hunt target priority
- Test if we need to check if cannon exists on another world.... What happens if the player world hops? Does %mcannon_coord reset? I'll need to research this. Current osrs it just lets you grab another cannon instantly, old cannon becomes non-functional because its *no longer* your cannon


thesuddensilent:
lol    I've never heard of that or seen anything like that.
test on staging world if you can steal other peoples cannons with world hopping (authentic: https://youtu.be/yTiq3TRZrys)

tannerdino:
> lol I've never heard of that or seen anything like that. test on staging world if you can steal other peoples cannons with world hopping (authentic: https://youtu.be/yTiq3TRZrys)

Also this video: https://www.youtube.com/watch?v=5wehnFqQVeI

PR #1497, opened 2025-03-18
fix(content): riff header on top of attack1.mid

ultraviolet-jordan:
I checked and it's the only one like this. It has a RIFF header on top of the MThd header.
image


lesleyrs:
Client-TS skips it already https://github.com/misterhat/tinymidipcm/compare/2642472d65938ab2451cae0799671783a66c5bd2...20c722b27d18256e3981335b0548134e82379c04

ultraviolet-jordan:
> Client-TS skips it already [misterhat/tinymidipcm@2642472...20c722b](https://github.com/misterhat/tinymidipcm/compare/2642472d65938ab2451cae0799671783a66c5bd2...20c722b27d18256e3981335b0548134e82379c04)

Not really the solution though is it. There shouldn’t be 2 file format headers on this single file. Because the header is there, it is not considered to be a valid MID. I highly doubt this is from Jagex unmodified.

Pazaz:
@ultraviolet-jordan it is indeed original. In 317-era caches it's archive 3 group 24. Microsoft's MIDI synth reads RIFF files no problem, I'm not sure why this file specifically has it

Pazaz:
317 cache: https://archive.openrs2.org/caches/runescape/876/flat-file.tar.gz
`3/24.dat` -> gunzip -> 
image

PR #1500, opened 2025-03-18
feat(client): Add auto client sizing

puremana:
Added auto sizing because often the 2x or 3x sizing doesn't fit right for me, and the fullscreen means I can't use other tabs when I am playing on a single monitor.

Below is a demo.

https://github.com/user-attachments/assets/43574e09-5e6f-4417-ab1b-d364af447b65

I have added a 120px height margin for the max width of the client, because in the real client there is the top navigation bar. This might need to change slightly for the real thing.

This is easy to change, it's just this line here

`let maxWidth = ((window.innerHeight - 120) / 532) * 789; `

Make that 140 instead of 120 or etc.

### DIFF
--- a/view/client.ejs
+++ b/view/client.ejs
@@ -143,6 +143,28 @@
             return size;
         }
  
+        function toggleAutoSizing() {
+            let canvas = document.getElementById('canvas');
+            let autoSizing = localStorage.getItem('autoSizing') !== 'true';
+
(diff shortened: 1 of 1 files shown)


Collin-M:
Auto scaling + clicking 1x/2x/3x makes the ratio wonky
image

PR #1512, opened 2025-03-20
fix(engine): After changing type, Npcs will now revert themselves automatically

mark-b5:
Before we were having to manually revert Npcs after changing their type

Now, `npc_changetype(Npc type, int duration)` takes in a `duration` and the engine will handle reverting the Npc after the duration has elapsed

I've gone ahead and filled all the examples in runescript of the command with a placeholder `^max_32bit_int`, so the feature is dormant for the time being

### DIFF
--- a/data/src/scripts/macro events/scripts/fishing/macro_event_whirlpool.rs2
+++ b/data/src/scripts/macro events/scripts/fishing/macro_event_whirlpool.rs2
@@ -1,7 +1,7 @@
 [proc,macro_event_whirlpool_spawn]
 %macro_event = ^no_macro_event;
 
-npc_changetype(npc_param(whirlpool));
+npc_changetype(npc_param(whirlpool), ^max_32bit_int);
 sound_synth(whirlpool, 0, 0);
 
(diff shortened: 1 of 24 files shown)


data/src/scripts/macro events/scripts/fishing/macro_event_whirlpool.rs2
```diff
@@ -1,7 +1,7 @@
 [proc,macro_event_whirlpool_spawn]
 %macro_event = ^no_macro_event;
 
-npc_changetype(npc_param(whirlpool));
+npc_changetype(npc_param(whirlpool), ^max_32bit_int);
```
tannerdino, line 4: Old videos show this as 60 ticks, previous implementation was wrong

mark-b5: Can be done in a future merge. This PR isn't modifying current behavior

tannerdino: Yeah sorry just noticed right after I made that comment!

Issue #1509, opened 2025-03-20
Rogues are aggressive (also making them impossible to pickpocket)


uldahlalex:
I'm trying to obtain a lockpick (possible pickpocket reward from rogues), but they are aggressive.

puremana:
What if you wait 10 minutes for the rogues to de-aggro?

PR #1518, opened 2025-03-22
fix: Adds aggression exemption for rogues

StealthyApe:
https://github.com/2004Scape/Server/issues/1509

addressing this issue here, it looks like they've provided proof on discord.

The issue seems to  be everything in the wilderness is treated as aggressive so somebody at the time must have written a carve out for them?

### DIFF
--- a/data/src/scripts/_unpack/all.npc
+++ b/data/src/scripts/_unpack/all.npc
@@ -4919,8 +4919,6 @@ param=damagetype,^stab_style
 param=attack_anim,human_sword_stab
 param=defend_anim,human_sword_defend2
 param=attack_sound,hacksword_stab
-huntmode=cowardly
-huntrange=1
 // Stats match osrs version but our version Vislvl is different.
 // https://raw.githubusercontent.com/Joshua-F/osrs-dumps/refs/heads/master/config/dump.npc npc_526
 


stomponbabyheads:
would you link said proof?

dahun0:
> would you link said proof?

I'm the one who posted about this on the discord, https://youtu.be/H9-qfBQRCqw?list=PLn23LiLYLb1Yiwi3sZJfeyANi_ndj67E8&t=249. You can see a rogue walking passed a player on the other side of the wall here. Further note is this is a monster you're supposed to be able to thieve and it doesn't make any sense for it to be aggressive. It's not aggressive in rsc preservation, rs3 or osrs.

ultraviolet-jordan:
And according to the wiki:
https://oldschool.runescape.wiki/w/Rogue#Level_15

*Unlike many other NPCs found in the Wilderness, rogues are not hostile and will only attack players if provoked or if they try to loot the [chests](https://oldschool.runescape.wiki/w/Chest_(Rogues%27_Castle)) within their [castle](https://oldschool.runescape.wiki/w/Rogues%27_Castle).*

So instead of hard checking for this npc on the engine, I think you should just remove these:
```ts
huntmode=cowardly
huntrange=1
```

From the "rogue" npc config:
https://github.com/2004Scape/Server/blob/65ef4d739c3cd8d0bde90118605fa6b85b004d6d/data/src/scripts/_unpack/all.npc#L4922C1-L4923C12

So this will cause them to not try to hunt players at all.

StealthyApe:
I wasn't too sure if the aggression was tied to the standard double combat level, if that isn't the case I'll modify the npc config instead.

StealthyApe:
Changed to the suggestion of [ultraviolet-jordan](https://github.com/ultraviolet-jordan) as it appears no aggression is expected at all

ultraviolet-jordan:
https://github.com/user-attachments/assets/b771a790-b6e6-48c5-a534-fbf1c1fb490a

PR #1525, opened 2025-03-23
fix(content): More authentic fishing spot points

thesuddensilent:
-changed some fishing spot movement points.
-The last named audio files have been moved to the folders they belong to.


tannerdino:
Fishing spot changes seem good to me

PR #1529, opened 2025-03-23
fix(engine): default visibility to 'hard' for staff

xVye:


ultraviolet-jordan:
fwiw I believe even they are visible by default based on this:
https://youtu.be/JGB_xYrpwsI?t=2789

xVye:
Thanks for the quick input!
We've had issues on live when disconnected with a reconnect to be set back to default(0).

ultraviolet-jordan:
I don't know what should happen on a reconnect, if I had to guess, it probably gets reset for them.

It could be made to persist through logout by making it use a varp and setting visibility based on that, not sure though just spitting ideas.

xVye:
Thank you, I'll take a look there!

ultraviolet-jordan:
The import for `Visibility` is gonna be different if you pull in changes.

xVye:
I've drafted this PR for now, as I'm trying to explore a better way of keeping the visibility level on reconnects, rather than on login.

PR #1531, opened 2025-03-23
feat(engine): prevent teleporting to invalid areas of the map

ultraviolet-jordan:
https://youtu.be/NN5O08mNxTg?t=1170

https://youtu.be/_a-y7BHtdV8?t=120


mark-b5:
Should be the case for Npcs too

ultraviolet-jordan:
> Should be the case for Npcs too

I have updated it so it checks for both players & npcs.

PR #1541, opened 2025-03-24
fix(content): prevent picking up multiple mage arena capes

tannerdino:
refactors @pickup_obj label as well


ultraviolet-jordan:
The reason `pickup_obj` passes in those args is bcuz we were taught that it is generally safer to call those commands from the trigger where everything starts and pass in all in.

mark-b5:
> The reason `pickup_obj` passes in those args is bcuz we were taught that it is generally safer to call those commands from the trigger where everything starts and pass in all in. There are many procs which technically don't need args.


@ultraviolet-jordan 

I think it was significantly *less* safe when it had arguments because the arguments didn't necessarily match the active Obj. The proc would do all these safety checks against the arguments you pass in and then just,..... pickup the active Obj which could be completely irrelevant to the safety checks that were just performed

ultraviolet-jordan:
> > The reason `pickup_obj` passes in those args is bcuz we were taught that it is generally safer to call those commands from the trigger where everything starts and pass in all in. There are many procs which technically don't need args.
> 
> @ultraviolet-jordan
> 
> I think it was significantly _less_ safe when it had arguments because the arguments didn't necessarily match the active Obj. The proc would do all these safety checks against the arguments you pass in and then just,..... pickup the active Obj which could be completely irrelevant to the safety checks that were just performed
> 
> ```ts
> [label,pickup_obj_floor](obj $obj, int $amount)
> if (~pickup_obj_check_for_space($obj, $amount) = false) {
>     return;
> }
> anim(null, 0);
> obj_takeitem(inv); // Doesn't necessarily have anything to do with $obj or $amount
> sound_synth(pick2, 0, 0);
> ```

I recognize your theortetical however you would have to change your active obj in the middle of the proc before you did the pickup.

mark-b5:
> change your active obj in the middle of the proc

I'm belaboring the point so I can procrastinate my work day but the issue is that our code calls `@pickup_obj(obj_coord, obj_type, obj_count);` 23 separate times. Passing literally any params to this label other than `obj_coord, obj_type, obj_count` would cause erratic behavior. These 3 arguments only introduce safety risk and don't offer any safety advantage

ultraviolet-jordan:
Definitely do not look into the last_int command then.

PR #1555, opened 2025-03-26
fix(content): unequipping items by conflict should be moved to the first slot of the inv

tannerdino:
This doesnt match osrs. But it matches old videos such as:

### [Example 1:](https://youtu.be/Hv3Ws6MjMQo?list=PLn23LiLYLb1Yiwi3sZJfeyANi_ndj67E8&t=71)

https://github.com/user-attachments/assets/9203fd7b-3a99-46b0-a2ec-3c856f6998e2

### [Example 2:](https://www.youtube.com/watch?v=4MEQkHvFJBc&list=PLn23LiLYLb1Yiwi3sZJfeyANi_ndj67E8&t=41s)

https://github.com/user-attachments/assets/948ada56-d55d-4a0b-988d-4a124be7eb22

I also went ahead and refactored the equip code with this pr


ultraviolet-jordan:
Rewriting the code is going to introduce risk, I highly recommend not doing this and just do the fix that you intended to do.

tannerdino:
> Rewriting the code is going to introduce risk, I highly recommend not doing this and just do the fix that you intended to do. I can already guarantee 3 follow up PR’s after this one if this goes in. We have done testing for a few years now and to go in 2 days before full release with rewrites is not recommended.

Its just easier to rewrite it, I think it matches all edge cases. The only way to match it with the vids is either with an inv_movetoslot command that doesnt swap slots, or a temp inv with a size of 1.

Inv with a size of 1 makes a lot of sense with my implementation, since it checks if unequipping conflicting equipment has enough space *if the obj to equip isnt in your inventory*. I dont really want to attempt doing the temp inv implementation with our current code

tannerdino:
List of tests:
- Items unequipped by conflict goes to the first available slot of the inventory

https://github.com/user-attachments/assets/bd8f215e-1a16-43e9-88e2-e8725657588d

- Wearing a 2h and equipping a shield unequips the 2h and vise versa

https://github.com/user-attachments/assets/d668b04a-e053-4b06-8f12-d788ac764889

- Stackable items unequipped by conflict will fail inv space check if the stackable item will reach above the stack limit (need confirmation?)

https://github.com/user-attachments/assets/889ccfaa-77eb-4066-a7f8-5d530a0d1bab

- Stackable items equipped will add as many as it possibly can to the stack limit

https://github.com/user-attachments/assets/2d3840dd-6485-4402-b90c-f3ef955ec391

tannerdino:
One thing I had to do in order to get this code to work is 
```java
if (inv_total(worn, $equip_obj) > 0 & oc_stackable($equip_obj) = true) {
    inv_moveitem(equipitem, worn, $equip_obj, $equip_obj_num);
} else {
    inv_movetoslot(equipitem, worn, 0, oc_wearpos($equip_obj)); // equip item :D
}
```
Because inv_movetoslot would *always* swap the two slots, even if they can combine... Not sure if this is correct? Do we have examples of same obj stackable items swapping in osrs?

Pazaz:
Inv 96 had its use found - see if you can find a different inv that matches your use (all were renamed)

tannerdino:
Redid the tests with my new code:

- Items unequipped by conflict goes to the first available slot of the inventory

https://github.com/user-attachments/assets/6a775c88-0441-4702-b8c0-62c2186df8db


- Wearing a 2h and equipping a shield unequips the 2h and vise versa

https://github.com/user-attachments/assets/88be5050-390b-4a81-aedc-5596fd4d03d4


- Stackable items unequipped by conflict will fail inv space check if the stackable item will reach above the stack limit (need confirmation?)

https://github.com/user-attachments/assets/841cffba-6ad1-4eb1-bca1-eb0b574f1723


- Stackable items equipped will add as many as it possibly can to the stack limit


https://github.com/user-attachments/assets/8c321409-5c02-4660-bc55-f07c57999718

ultraviolet-jordan:
Have you tried testing with something like a `plague_jacket`? And also with stackable weapon? And any combinations of all of your examples + my examples with a full inv?

tannerdino:
https://github.com/2004Scape/Server/blob/c769fbe384563413a39232684c42efd780e7017b/data/src/scripts/player/scripts/equip.rs2#L342-L347

So I do this to replicate the behavior in the old videos. My theory is that they did `inv_movefromslot(inv, inv $inv_slot)` to put the swapped item into the first slot of the inventory, as... a way to stack stackable items?

Removing the `inv_movefromslot` would give you the exact behavior as seen in https://youtu.be/EYXp7GMFWEw, https://youtu.be/DwzDt3yw6k4. So perhaps thats what the late 2006 change was for the arrow bug.

tannerdino:
Are these tests any good?

- Equipping 2h with a full inv, whilst having shield + knives equipped


https://github.com/user-attachments/assets/61460d50-f761-4fe7-8a85-25bb3e1751b4


- Equipping 2h with full inv and knives, whilst having shield + knives equipped

https://github.com/user-attachments/assets/f8120aa7-e6cd-4a99-b7cb-98e16c89cb58



- Plague jacket:

https://github.com/user-attachments/assets/1b25ce99-201f-4952-b580-0c010b18e201


- Plague jacket with full inv:

https://github.com/user-attachments/assets/3e223e93-34b8-4a20-9ca6-4bb930e7731b

tannerdino:
What was that 2h + shield bug? Was it just this:


https://github.com/user-attachments/assets/5de6732e-e557-438d-aec8-8aac3828f269

ultraviolet-jordan:
Stackable weapon at max stack and stuff. You quite literally MUST think of every possible edgecase scenario you can possibly muster.

tannerdino:
All seems to match intended behavior 😎 


https://github.com/user-attachments/assets/f17e63f6-9772-408f-8c20-eb6583268bb0

ultraviolet-jordan:
Use a plague jacket now to conflict max stack weapons

ultraviolet-jordan:
Plague jacket doesnt conflict weapons does it? Nvm

ultraviolet-jordan:
I think ur gucci then I hope I'm not missing something tho I had all those requirements in my head 2 yrs ago. I wrote a bunch of stuff down in the code comments at the time but ye.

tannerdino:
Ill try to think of more edge cases

thesuddensilent:
What happens if you try to sell something to a general store if there is already a max stack of that item?

ultraviolet-jordan:
Or anything with future items that we necessarily won't run into right now. I remember chins being brought up a few times during the development. But I think chins should be covered here already tho.

ultraviolet-jordan:
> What happens if you try to sell something to a general store if there is already a max stack of that item?

It's already covered. This PR is only covering this equipping rewrite.

PR #1559, opened 2025-03-27
feat: account for -1 ticks in duration validation

MattGould1:
Hey recently we changed how ticks work in `trackLocObj` but we didn't update our validations for loc/objs. This meant we could end up adding loc/objs for a `0` tick duration.

https://github.com/2004Scape/Server/blob/main/src/engine/World.ts#L1401

### DIFF
--- a/src/engine/zone/Zone.ts
+++ b/src/engine/zone/Zone.ts
@@ -99,30 +99,28 @@ export default class Zone {
 
     tick(tick: number): void {
         for (const obj of this.getAllObjsUnsafe()) {
-            if (!obj.updateLifeCycle(tick) || obj.lastLifecycleTick === tick) {
-                continue;
-            }
-            if (obj.lifecycle === EntityLifeCycle.DESPAWN) {
-                if (obj.reveal !== -1) {
-                    World.revealObj(obj);
-                } else {
-                    World.removeObj(obj, 0);
+            if (obj.lifecycleTick > -1 && obj.lifecycleTick <= tick) {
+                if (obj.lifecycle === EntityLifeCycle.DESPAWN) {
+                    if (obj.reveal !== -1) {
+                        World.revealObj(obj);
+                    } else {
+                        World.removeObj(obj, 0);
+                    }
+                } else if (obj.lifecycle === EntityLifeCycle.RESPAWN) {
+                    World.addObj(obj, Obj.NO_RECEIVER, 0);
                 }
-            } else if (obj.lifecycle === EntityLifeCycle.RESPAWN) {
-                World.addObj(obj, Obj.NO_RECEIVER, 0);
             }
         }
 
         for (const loc of this.getAllLocsUnsafe()) {
-            if (!loc.updateLifeCycle(tick) || loc.lastLifecycleTick === tick) {
-                continue;
-            }
-            if (loc.lifecycle === EntityLifeCycle.DESPAWN) {
-                World.removeLoc(loc, 0);
-            } else if (loc.lifecycle === EntityLifeCycle.RESPAWN && loc.isChanged()) {
-                World.revertLoc(loc);
-            } else if (loc.lifecycle === EntityLifeCycle.RESPAWN && !loc.isActive) {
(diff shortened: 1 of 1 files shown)


mark-b5:
Are we sure Jagex wouldn't allow a 1 tick duration in their scripts? Intuitively I think their engine might just allow it

mark-b5:
https://github.com/user-attachments/assets/24937e87-2378-459f-b97c-6f13dde593d0

In fact, our npc_change function uses the same duration validation, and from this OSRS clip we know that we need to allow a 1 tick duration

mark-b5:
^ Hmmmmmmm, that might be a duration of `2` in osrs actually,,,,,,,

MattGould1:
That could be, I think something is a little off then because:

`loc_add(coord, loc, angle, shape, 1);`
`entity.setLifeCycle(this.currentTick + duration - 1);`

I'm not setting the loc for 1 tick here, its being set for 0 ticks

I think then in Entity.ts we do this

```
    updateLifeCycle(tick: number): boolean {
        return this.lifecycleTick === tick && this.lifecycle !== EntityLifeCycle.FOREVER;
    }
```

then in Zone.ts we do this

```
        for (const obj of this.getAllObjsUnsafe()) {
            if (!obj.updateLifeCycle(tick) || obj.lastLifecycleTick === tick) {
                continue;
            }
```

We end up adding the loc/obj forever I think by accident

MattGould1:
(I have no idea about the authenticity here, I'd imagine they did adjust the tick rates like we've done but I think our validation needs to account for that?)

mark-b5:
So there's a core engine mismatch between us and OSRS atm

We store all of our `lifecycle` events as `World.currentTick + duration` then compare `lifecycletick === World.currentTick`

I'm pretty much completely confident osrs stores all of theirs as just `lifecycletick = duration` and then every tick they do `--lifecycletick <= 0`. Essentially, instead of scheduling for a specific tick, they just store a counter on the entity itself and then decrement it every tick. This causes a series of small mismatches between us and them

The `duration-1` thing is a bandaid hack for the time being to get our timing to match theirs without completely rewriting the system,.... but the whole thing is due for a rewrite

I think for the time being we could put an extra safety check in `World.tracklocobj` to make sure that duration is some point in the future

MattGould1:
does it make sense to add that check in `World.tracklocobj` and not in `Entity.updateLifeCycle`? I might misunderstand what the solution needs to be here but I think we could do

```
    updateLifeCycle(tick: number): boolean {
        return this.lifecycleTick  <= tick && this.lifecycle !== EntityLifeCycle.FOREVER;
    }
 ```
 
 ?

mark-b5:
{C2C22D0F-F6E0-46FE-99BA-91FDBA23D3AF}

Hmmm that looks good to me, since this is the only place that method is being called

MattGould1:
sweet I'll change this and go pass out and wait for some feedback in the morning (if it exists)

mark-b5:
> sweet I'll change this and go pass out and wait for some feedback in the morning (if it exists)

Wait it might have funny behavior because I don't think we clear entity.lifecycletick when the events happen. In other words, I think it would cause our lifecycle events to happen more often than they should because `lifecycletick <= world.currenttick` will be true forever and ever

It should be something like

`if (npc.lifecycleTick > -1 && npc.lifecycleTick < this.currentTick) {`

and then make sure to set `lifecycleTick` to `-1` after a lifecycle event happens

MattGould1:
Hey I applied your suggestion to locs/objs and this is the behaviour https://streamable.com/l1509l

I have no idea about whether this is authentic however it does fix the problem I found yday:

1. If you add a loc/obj with a 1 tick duration, it's removed from the game world
2. Other actions in the game are one tickable and now so are adding loc/objs to the game world

mark-b5:
Looks good to me 

https://github.com/2004scape/Server/blob/main/data/src/scripts/_test/scripts/cheats/cheat_zone.rs2#L1

There's some test commands here if you want to make sure everything is sane

PR #1566, opened 2025-03-28
fix(content): pizza halves should appear in first slot after eating a full pizza

tannerdino:
todo: figure out if stuff like cakes should use this?


thesuddensilent:
probably almost all of them work in the same way.  but I can't say with 100% certainty.

PR #1587, opened 2025-03-30
fix(engine): Loc and Obj events are now handled in the correct order (fixed the gnome stronghold door)

mark-b5:
We need Loc and Obj events (specifically Loc) to be processed in the order they are initialized. Previously, they would process in the order that they exist within the Zone's LinkedList. This would give incorrect behavior when multiple events need to happen on the same tick. Specifically, gnome stronghold door was behaving erratically

Now the events are stored in a nice event queue LinkedList stored on World. This makes Loc and Obj events match RSProx output and osrs behavior. It also gets rid of the jank `duration -1` that existed inside of `trackLocObj`

### DIFF
--- a/src/engine/World.ts
+++ b/src/engine/World.ts
@@ -1550,16 +1550,8 @@ class World {
     }
 
     revealObj(obj: Obj): void {
-        // printDebug(`[World] revealObj => name: ${ObjType.get(obj.type).name}`);
-        // const duration: number = obj.reveal;
-        // const change: number = obj.lastChange;
         const zone: Zone = this.gameMap.getZone(obj.x, obj.z, obj.level);
         zone.revealObj(obj);
-        // objs next life cycle always starts from the last time they changed + the inputted duration.
-        // accounting for reveal time here.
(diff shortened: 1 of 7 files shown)


src/engine/World.ts
```diff
@@ -1490,75 +1510,77 @@ class World {
...
+        const zone: Zone = this.gameMap.getZone(obj.x, obj.z, obj.level);
         if (ObjType.get(obj.type).stackable && obj.lifecycle === EntityLifeCycle.DESPAWN) {
             const existing = this.getObjOfReceiver(obj.x, obj.z, obj.level, obj.type, receiver64);
             if (existing && existing.lifecycle === EntityLifeCycle.DESPAWN) {
                 const nextCount = obj.count + existing.count;
                 if (nextCount <= Inventory.STACK_LIMIT) {
-                    // if an obj of the same type exists and is stackable and have the same receiver, then we merge them.
-                    this.changeObj(existing, receiver64, nextCount);
+                    // If an obj of the same type exists and is stackable and have the same receiver, then we merge them.
+                    this.changeObj(existing, nextCount);
```
ultraviolet-jordan, line 1526: changeObj already tracks zone so you are doing double work.

mark-b5:
Added some more failsafes to avoid memory leaks

PR #1588, opened 2025-03-31
fix(content): add missing NIH to digsite altar bricks

xylphie:


Indio3:
did you confirm this on OSRS? a few interactions miss NIH in their default case

xylphie:
I did not. I'll ask the author if this was intentional since I don't have an account I can test this on

PR #1590, opened 2025-03-31
fix(content): Fixing lava eel fishing for loc

AmVoidGuy:
There was no increment on the action delay while fishing for lava eels from the loc object, so it would perpetually try to fish but never catch anything in the spawns in the lava maze.


tannerdino:
looks good

PR #1597, opened 2025-03-31
fix(content): only scatter tradables on genie fail

tannerdino:
There is no proof of how this piece of content worked. But its particularly buggy if we dont check for untradeables. Lots of untradeables have custom opheld5 triggers, inv_dropslot would skip the custom behavior in these triggers.

### DIFF
--- a/data/src/scripts/macro events/scripts/macro_events.rs2
+++ b/data/src/scripts/macro events/scripts/macro_events.rs2
@@ -131,8 +131,9 @@ session_log(^log_adventure, "Failed random event and had their items scattered")
 p_delay(1);
 def_int $i = 0;
 while ($i < inv_size(inv)) {
-    def_coord $coord = map_findsquare(coord, 0, 5, ^map_findsquare_lineofwalk); // 11x11 for now. todo: confirm how scattering works
-    if (inv_getobj(inv, $i) ! null) {
+    def_obj $obj = inv_getobj(inv, $i);
+    if ($obj ! null & oc_tradeable($obj) = true) {
+        def_coord $coord = map_findsquare(coord, 0, 5, ^map_findsquare_lineofwalk); // 11x11 for now. todo: confirm how scattering works
         inv_dropslot(inv, $coord, $i, 200);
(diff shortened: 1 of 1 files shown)


ultraviolet-jordan:
Why

tannerdino:
> Why

Typo in the title sorry!

ultraviolet-jordan:
Why only tradeables

tannerdino:
Some untradeable items arent supposed to be dropped (or have special behavior when dropped), and we really dont know how scattering even worked... or when it was removed

ultraviolet-jordan:
> Some untradeable items arent supposed to be dropped (or have special behavior when dropped), and we really dont know how scattering even worked... or when it was removed

Which untradeable items are not supposed to be dropped or have special behavior when dropped (basically I am asking why are you opening this PR plz use your keyboard)

Indio3:
> > Some untradeable items arent supposed to be dropped (or have special behavior when dropped), and we really dont know how scattering even worked... or when it was removed
> 
> Which untradeable items are not supposed to be dropped or have special behavior when dropped (basically I am asking why are you opening this PR plz use your keyboard)

almost all of the non default opheld5 scripts

tannerdino:
God capes, gnome ball, bloated toad, family crest gauntlets, ana in a barrel, pets, unidentified liquid, nitroglycerin, mixed chemicals, chemical compound, ghost skull, witches cat, stone plaque, tattered scroll, crumpled sccroll, zadimus corpse, sword pommel, crystal, bervirius notes, bone shard, bone beads, beads of the dead, bone key, rashilya corpse

I can revert this unconfirmed behavior back to the previously unconfirmed behavior if you want....

mark-b5:
> I can revert this unconfirmed behavior back to the previously unconfirmed behavior if you want....

I think he just wants an explanation in the OP

tannerdino:
Updated the desc

ultraviolet-jordan:
Was only looking for an explanation since these PRs are truly our own receipts as to why X does Y and is how I am able to go back and see. It hurts me when I need to look into something and there was no information, although we have been decent about it, I think we can do better.

PR #1600, opened 2025-03-31
fix(content): Fix grammar on magic tab.

EFabrizioGC:
This PR fixes the wrong grammar in the Fire wave mage spell description.


thesuddensilent:
The typo in question is in the cache itself, so it is authentic.

PR #1605, opened 2025-04-01
fix(content): remove "You've been frozen!" mes

tannerdino:
- No "You are frozen!" mes. I believe this mes comes from ancients later on

From [video](https://youtu.be/1A_AITzasxE?t=431):

https://github.com/user-attachments/assets/4645e0dd-b564-46fd-88f6-571f2bd867e2

A potential issue with our engine code though...

https://github.com/user-attachments/assets/b73c3efd-9bb5-4703-bbab-5e6aeb2a2d4a

Our walktrigger runs twice, whilst the video it runs once?

image
image

### DIFF
--- a/data/src/scripts/skill_combat/scripts/npc/npc_combat_magic.rs2
+++ b/data/src/scripts/skill_combat/scripts/npc/npc_combat_magic.rs2
@@ -55,7 +55,6 @@ if (~npc_player_hit_roll(^magic_style) = true) {
 
 // todo: dunno if this should be queued but protected access lol
 [queue,npc_freeze_player](int $duration)
-mes("You've been frozen!");
 %frozen = calc(map_clock + $duration + 1);
 walktrigger(frozen);
 
--- a/data/src/scripts/skill_combat/scripts/pvp/pvp_magic.rs2
+++ b/data/src/scripts/skill_combat/scripts/pvp/pvp_magic.rs2
@@ -241,8 +241,8 @@ if (inv_total(worn, gauntlets_of_chaos) > 0 & $spell >= ^wind_bolt & $spell <= ^
 }
 return($maxhit);
 
+// https://youtu.be/1A_AITzasxE?t=431
 [queue,pvp_freeze_player](int $duration)
-mes("You've been frozen!");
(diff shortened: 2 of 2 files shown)


tannerdino:
I think the engine issue is:

```ts
// Process walktrigger if there is waypoints
if (this.hasWaypoints() && this.canAccess()) {
    this.processWalktrigger();
}
```

This `this.hasWaypoints()` returns true if p_walk(coord) is ran

mark-b5:
```typescript
            // Recalc path
            this.pathToPathingTarget();

            // Process walktrigger if there is waypoints
            if (this.hasWaypoints() && this.canAccess()) {
                this.processWalktrigger();
            }
```

I think it's because a path recalc will happen immediately before this. Can you check osrs behavior on this? I thought I verified that osrs runs the walktrigger twice in this case

Maybe it's the case that path recalc was handled differently in 04

mark-b5:
Is there a clip of someone meleeing an opponent while stunned?

mark-b5:
```typescript
        // If there is a target and p_access is available, try to interact before movement
        if (this.target && this.canAccess()) {
            // Clear the interaction if target validation does not pass
            if (!this.validateTarget()) {
                this.clearInteraction();
                this.unsetMapFlag();
                return;
            }

            // Run the optrigger, but applayer3 should not run this
            if (!followOp) {
                this.processWalktrigger();
            }

            interacted = this.tryInteract(false);
        }

```

I'm wondering if the processwalktrigger here is a newer addition, and a clip of someone meleeing while stunned would show the behavior

tannerdino:
> Is there a clip of someone meleeing an opponent while stunned?

https://youtu.be/_NmFftkMm0I?t=81

tannerdino:
You can see from here and other parts of the video. It only shows once on melee attack (from moveclick packet handler), if within op distance.

https://github.com/user-attachments/assets/d533b0c7-3687-46cd-8735-73c49ba2b33a

If out of op distance, itll play once from moveclick and once in interactions


https://github.com/user-attachments/assets/f2bff04b-d816-4948-aa1f-f8148d6bab65

PR #1612, opened 2025-04-02
fix(content): ents were making trees walk throughable

tannerdino:
Relies on https://github.com/2004Scape/Server/pull/1587

### DIFF
--- a/data/src/scripts/macro events/configs/antimacro.npc
+++ b/data/src/scripts/macro events/configs/antimacro.npc
@@ -1278,7 +1278,6 @@ wanderrange=0
 moverestrict=nomove
 category=macro_event_ent
 param=ent_tree_loc,loc_1276
-param=next_loc_stage,loc_2639
 
 [macro_event_ent_tree2]
 vislevel=hide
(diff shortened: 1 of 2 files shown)


ultraviolet-jordan:
use `loc_2639` instead of `loc_83` also your original code was using a `loc_84` and the new code doesn't. Don't think there is a diff between them tho.

ultraviolet-jordan:
Oh I see you're going from not using `loc_2639`, what's wrong with it? It looks the same except 2x2

tannerdino:
The order you loc_del/loc_change and loc_add is rly sensitive it looks like. In the previous code, the first loc reverts before the next locs delete, which breaks the blocking I guess.

The excessive loc_83's was me playing it save with no overlapping locs, but looking back it probably doesn't matter at all.


Before:


https://github.com/user-attachments/assets/ae891f17-64de-4804-8549-ea3d26dce160



After:


https://github.com/user-attachments/assets/b68a024c-c383-4a83-b554-07e7fb043c44

tannerdino:
Btw the tree spawning before the ent disappears is authentic believe it or not!
- https://youtu.be/1vo3B5DEgIQ?list=PLn23LiLYLb1aYjPgadhUQGX61AeHWaT4y&t=25
- https://youtu.be/eGLAXnneCsg?list=PLn23LiLYLb1aYjPgadhUQGX61AeHWaT4y&t=260

ultraviolet-jordan:
Ah yeah that makes sense if so. And 1x1 or the 2x2 should functionally be the same I would think... I couldn't find a 3x3. But uh the only thing with using all the 1x1's is the script is doing extra work basically.

PR #1618, opened 2025-04-03
fix(content): steelskin prayer drain could exponentially stack

tannerdino:
oops!

Video of bug:

https://github.com/user-attachments/assets/6160419a-bca4-4207-9563-cb0057248b77


ultraviolet-jordan:
What about superhumanstrength

tannerdino:
Steelskin is the only one I noticed, I can double check every prayer again though. Just in case

tannerdino:
Heres me testing every prayer:

https://github.com/user-attachments/assets/0af686a5-cd6d-40fa-8401-8c6d5cbffc70

ultraviolet-jordan:
Ok it was probably fixed already. I was looking at the PR that changed prayers recently.

PR #1619, opened 2025-04-03
fix(content): clear pending action when unequipping items

tannerdino:
From [video](https://youtu.be/AVpqoWjsT1E?t=34):

https://github.com/user-attachments/assets/8a7d7268-e964-4f50-8fdb-a04d3533377b

Makes me question... Did every protected if_button/inv_button in the engine clear pending action and close interface?

### DIFF
--- a/data/src/scripts/player/scripts/equip.rs2
+++ b/data/src/scripts/player/scripts/equip.rs2
@@ -2,6 +2,7 @@
 
 [inv_button1,wornitems:wear]
 if_close;
+p_clearpendingaction;
 ~unequip(last_slot);
 
 [proc,equip](int $slot)


mark-b5:
Yeah I suspect it's just an engine thing

tannerdino:
> Yeah I suspect it's just an engine thing

Yeah we use the p_clearpendingaction command, which seems like it'd be an obscure command they wouldn't really use often.

tannerdino:
Not sure if its the same thing but early videos show the special attack button *not* clearing interactions (matches osrs)

tannerdino:
Heres another video, of setting autocast clearing interactions: https://youtu.be/P_QzMzpaA04?t=12

PR #1625, opened 2025-04-03
fix(content): players shouldnt be able to duel on f2p worlds

tannerdino:
You can duel players in this area, which are accessible in f2p. Also missing return for dueling players outside the arena

image


thesuddensilent:
It's very likely that part 52,51 of the map is entirely p2p only, but I'm not entirely sure.

ultraviolet-jordan:
if it helps, this is what the client thinks is the duel arena area to show "Duel-with"

`x >= 3328 && x <= 3393 && z >= 3203 && z <= 3325`

thesuddensilent:
Is that duel option area defined in the cache/client itself?

tannerdino:
> if it helps, this is what the client thinks is the duel arena area to show "Duel-with"
> 
> `x >= 3328 && x <= 3393 && z >= 3203 && z <= 3325`

Oh nice, matches my inzone check perfectly

thesuddensilent:
From what I've watched on YouTube videos about the duel arena, the area defined as the duel area is super weird.
I don't think it even works according to the 8x8 zones.
https://www.youtube.com/watch?v=G-OB3K5pYL8
https://www.youtube.com/watch?v=BXGm3rTjmhw

mark-b5:
Probably changed after they introduced `area`s

PR #1626, opened 2025-04-04
fix(engine): Npcs and Players now have symmetric lines of sight (fixes 1-way safespots)

mark-b5:
It's always been the case in OSRS that Npcs will calculate their line of sight backwards when attacking a player. I.e., line of sight will be calced Player->Npc for both Entities. This will be pretty crucial for stuff like fight caves in 2005

## Previous behavior 
Npc can attack me from this position but I cannot attack it
image

## Updated behavior 
Npc cannot attack me and I cannot attack it. We share the same line of sight calculation between us

{3364AB58-D84D-47EB-AA6F-D5EC326C0C06}

### DIFF
--- a/src/engine/entity/PathingEntity.ts
+++ b/src/engine/entity/PathingEntity.ts
@@ -396,7 +396,7 @@ export default abstract class PathingEntity extends Entity {
             // you are not within ap distance of pathing entity if you are underneath it.
             return false;
         }
-        // Los is always calculated Player -> Npc, so that pvm interactions are symmetrical
+        // Los for Npcs is always calculated backwards for all Entity types (tested Player and Npc)
         if (this instanceof Npc) {
             return CoordGrid.distanceTo(this, target) <= range && isApproached(this.level, target.x, target.z, this.x, this.z, target.width, target.length, this.width, this.length);
         }


src/engine/entity/PathingEntity.ts
```diff
@@ -396,6 +396,10 @@ export default abstract class PathingEntity extends Entity {
             // you are not within ap distance of pathing entity if you are underneath it.
             return false;
         }
+        // Los is always calculated Player -> Npc, so that pvm interactions are symmetrical
+        if (this instanceof Npc && target instanceof Player) {
```
ultraviolet-jordan, line 400: can remove this `&& target instanceof Player`

mark-b5: Why? We only want to reverse Los direction for Npc interacting with a Player specifically. This conditional with a comment is the cleanest way I could imagine to accomplish this

ultraviolet-jordan: Do you know if Npc targetting Npc is the same way

mark-b5: I know that Npc -> Obj is normal

mark-b5: https://gyazo.com/aea0e5ca8d8a99b423622693335db833

I found an old clip of the box trap having los on the chin here, meaning it's calcing in the normal direction (box trap -> chin)

mark-b5: https://github.com/user-attachments/assets/e863ed1f-b0b8-4b73-8df9-acb0009046b9

https://github.com/user-attachments/assets/8d3d2ed7-c7cf-43ba-9bc6-2b70278f857d


Sooo these 2 clips show 2 things

Hunt los direction is gnome->warrior (normal)

But after gnome targets it can't attack (line of sight direction is gnome<-warrior)

So I think you're right. 

As for penance runners and varrock sweeper, it's only *hunt* line of sight that we know about. Their actual interaction isn't an ap

So to conclude:
1. `ap` lineofsight is backwards for Npcs targeting any entity
2. hunt lineofsight is backwards for Npcs hunting a player (tested)
3. hunt lineofsight is normal for Npcs targeting everything else (tested)

ultraviolet-jordan: Nice work big dog

PR #1627, opened 2025-04-05
fix(content): clue typo, zamorak wizard check, clue drops

Indio3:
- fix typo in Vigroy dialogue
- Change check for zamorak wizard to use huntall
- change droprate of hard clues from hellhounds to 1/64, add medium clues to tribesman, easy clues to thugs


tannerdino:
If you're changing clues you should probably remove the likely inauthentic if_close

tannerdino:
Sorry for clue drops, the if_close for that in osrs is related to combat achievements right

Indio3:
yeah we came to the conclusion it was CA related rather than a clue change

PR #1630, opened 2025-04-06
fix(content): no longer reopen puzzlebox when moving a piece

AmVoidGuy:


ultraviolet-jordan:
We can't change the .if files. Those are original from Jagex themselves.

AmVoidGuy:
Ah, I see. Can close this one then

ultraviolet-jordan:
@Indio3 Do you have any thoughts?

Do you need a way to prevent this though or maybe it’s legit for it to operate like this? Maybe we don’t have enough proof of anything.

AmVoidGuy:
It doesn't seem legit but that probably isn't what they do for it. Our ophelds currently call clearPendingAction which close interfaces. I'm not sure what the correct solution would be though. I see videos from 2006 where it doesn't close during puzzles but will need to look around more. In current Osrs it seems like ophelds don't close your open interface in general?

ultraviolet-jordan:
> It doesn't seem legit but that probably isn't what they do for it. Our ophelds currently call clearPendingAction which close interfaces. I'm not sure what the correct solution would be though. I see videos from 2006 where it doesn't close during puzzles but will need to look around more. In current Osrs it seems like ophelds don't close your open interface in general?

Yeah the main "problem" is they've had big interface system reworks over the years. And OSRS client today uses a custom language called "ClientScript" which handles a ton of interface stuff client sided automatically.

But uh........ the most obvious thing here is the `interactable` property on these couple interfaces in the game.
`trail_puzzle.if`
`inventory.if`

So *a* fix could be to wrap the `player.clearPendingAction();` with that check or something...

```ts
if (!com.interactable) {
     player.clearPendingAction();
}
```

tannerdino:
Is it possible that the puzzle interface was an overlay?

Indio3:
it's a modal on OSRS so I don't think so... I think they probably had a check before running `clearPendingAction`. It can't check on `interactable` or any property they share though since this wouldn't `clearPendingAction` for opheld on the inventory (which it is supposed to do). It should be checking on a property they don't share or checking if the com is not in the players current main modal

AmVoidGuy:
Checking if it's in the players current main modal or not seems to be the most practical solution for now. The only other properties they don't share are draggable and usable. It doesn't seem like either of those would make sense and a hardcoded check for the interface com name or id sounds bad. 

 It would work with our current use cases at least.

PR #1631, opened 2025-04-06
fix(content): bool behavior for Fishing Contest & Merlin's Crystal

Collin-M:
Not an ideal fix for Fishing Contest, but trying the varp check `! true` still doesn't work properly at hemenster.

`! true` does work as expected during Merlin's to get the wax

Reported on Discord by .starkwolf & .1174i
https://discord.com/channels/953326730632904844/1358296528883679244

### DIFF
--- a/data/src/scripts/areas/area_seers/scripts/hemenster/bonzo.rs2
+++ b/data/src/scripts/areas/area_seers/scripts/hemenster/bonzo.rs2
@@ -37,7 +37,7 @@ if($option = 1) {
 
 [proc,bonzo_set_places]
 ~chatnpc("<p,neutral>Ok, nearly everyone is in their place already.|You fish in the spot by the willow tree,|and the Sinister Stranger, you fish by the pipes.");
-if(%hemenster_pipe_stashed = true) {
+if(%hemenster_pipe_stashed ! null & %hemenster_pipe_stashed = true) {
     ~move_hemenster_pipe;
     return;
 }
--- a/data/src/scripts/areas/area_white_wolf_mountain/scripts/mountain_dwarf.rs2
+++ b/data/src/scripts/areas/area_white_wolf_mountain/scripts/mountain_dwarf.rs2
@@ -67,7 +67,6 @@ if($option = 1) {
         ~chatplayer("<p,happy>Fortunately I'm alright at fishing!");
         ~chatnpc("<p,happy>Okay, I entrust you with our competition pass.|Go to Hemenster and do us proud!");
         %fishingcompo_progress = ^fishingcompo_started;
-        %hemenster_pipe_stashed = false;
         ~send_quest_progress(questlist:fishingcompo, %fishingcompo_progress, ^fishingcompo_complete);
         inv_add(inv, fishing_pass, 1);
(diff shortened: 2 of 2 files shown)


AmVoidGuy:
image

Is there a reason we don't instantiate boolean varps to false? They are being set to -1 (null) by default when creating a new player. Seems like we assume in the code they should be false (0) by default?

Indio3:
fwiw, boolean varps did default to 0 (false) when both uses were written, I think it was changed to match what jagex does

Indio3:
I don't think you need to change the hemenster script right? It checks = true to move the stranger when it defaults to null, and won't be set until the garlic is stashed

Collin-M:
> I don't think you need to change the hemenster script right? It checks = true to move the stranger when it defaults to null, and won't be set until the garlic is stashed

[if(%hemenster_pipe_stashed = true) {](https://github.com/2004Scape/Server/blob/910f1910c1f489503e3885cebcff62c75f45b1a1/data/src/scripts/areas/area_seers/scripts/hemenster/bonzo.rs2#L40)
I've tried this with `! false` as well and the eval still returns true and `~move_hemenser_pipe` is triggered regardless of putting garlic in the pipe or not.

Collin-M:
It does need to be changed. Right now with `(%hemenster_pipe_stashed = true)` evals to true even if you haven't put garlic in the pipe (-1). You can complete this quest without garlic in the live game.

image
image

Indio3:
Oh that's odd... what if you try doing `%hemenster_pipe_stashed ! null & %hemenster_pipe_stashed = true`?

Collin-M:
image
image

That works. Is that a better solution you'd prefer instead of setting to false at quest start?

Indio3:
Yes, that way it wont be broken for any existing users who have started the quest w/out finishing

ultraviolet-jordan:
I brought this up to someone the other day, the variable system works but there's some duplicated code, and it stores variables in separate collections when it could just be 1. And the "default value" code kind of sucks. And yes I am pretty sure `boolean` should default to 0. 

Here is RuneScape 3 client with all of their types.
https://github.com/LostCityRS/RS910/blob/471d8354bc334b14854b1ffc1872c126b77fd5f4/client/src/main/java/com/jagex/game/config/vartype/constants/ScriptVarType.java#L11

```java
public ScriptVarType(int serialID, char legacyChar, BaseVarType baseType, Object defaultValue, ScriptVarProperty[] arg4) {

}
```

Collin-M:
> My only concern is introducing risk. Fixing this will _MOST LIKELY_ be followed up with a dozen other PR's.

fea608363142080259054485c294303c695960fb for now (temp fix) then we can work out this default value issue and affected varps during next major revision change (245)?

ultraviolet-jordan:
> > My only concern is introducing risk. Fixing this will _MOST LIKELY_ be followed up with a dozen other PR's.
> 
> [fea6083](https://github.com/2004Scape/Server/commit/fea608363142080259054485c294303c695960fb) for now (temp fix) then we can work out this default value issue and affected varps during next major revision change (245)?

Yeah it's chill I'm just making things known

PR #1013, opened 2024-11-20
feat(content): Legends Quest

noted-shark:
- [ ] Quest start
- [x] Chop jungle (might need tweaking, doesn't feel quite right)
- [x] Mapping
- [x] Bullroarer
- [x] Gujuo
- [x] Golden bowl
- [x] Ungadulu
- [x] Cave obstacles
- [x] Smell puzzle
- [ ] Gem puzzle (WIP)
- [ ] Fire wall (WIP)
- [ ] Book of binding (WIP)
- [ ] Nezikchened part 1 (WIP)
- [ ] Orb door
- [x] Bravery potion
- [ ] Winch
- [ ] Strange barrels
- [ ] Viyeldi
- [ ] Heart puzzle
- [ ] Boulders
- [ ] Nezikchened part 2
- [ ] Yommi tree
- [ ] Totem creation
- [ ] Nezikchened part 3
- [ ] Radimus finish & reward

PR #1637, opened 2025-04-07
feat(content): Legends quest

Indio3:
builds off https://github.com/2004Scape/Server/pull/1013


ultraviolet-jordan:
I have this mesbox if you need it:

image

Indio3:
already got this one, just noticed we're missing one of the mesboxes before that though so will add

data/src/scripts/quests/quest_legends/scripts/legends_boulder.rs2
```diff
@@ -0,0 +1,82 @@
+[ai_spawn,legends_boulder]
+%npc_start_coord = npc_coord;
```
tannerdino, line 2: I think you can put this in the global ai_spawn trigger I think. Random events in mid-osrs, would despawn if they teleported too far from their spawn coord I think? Could be an engine command... But hard to say!

ultraviolet-jordan: the boulder should have block walk on as well if it doesn't already. just spitting my brain.

thesuddensilent:
gallery_113_51_1099985217

You are probably missing this picture.

Indio3:
> gallery_113_51_1099985217
> 
> You are probably missing this picture.

good find, the formatting is slightly different to what we have

PR #1638, opened 2025-04-07
fix(engine): Ground item despawn time is sped up across the board by 100 ticks

mark-b5:
Previously we were adding 100 ticks onto `duration` when dropping items on the ground

We checked the available runescript leaks that we have and this didn't match observable osrs behavior (npc drops last a total of 200 ticks, not 200 +100)

image

It looks like (almost) all of our content was written with the assumption that `duration` is 1:1 with despawn time, so we don't need to do a bunch of content changes to match new engine behavior. I did change tables specifically to match new behavior though

### DIFF
--- a/data/src/scripts/general_use/scripts/tables.rs2
+++ b/data/src/scripts/general_use/scripts/tables.rs2
@@ -38,7 +38,7 @@ if ($x > 0) {
     $obj_coord = movecoord(coord, -1, 0, 0); // player is east of table
 }
 
-inv_dropslot(inv, $obj_coord, last_useslot, 900); // 1 min to be visible, 9 min to disappear
+inv_dropslot(inv, $obj_coord, last_useslot, 1000); // 1 min to be visible, 9 min to disappear
 
 [oplocu,_unusable_table]
 mes("The table appears to be in use.");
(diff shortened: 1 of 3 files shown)


ultraviolet-jordan:
Are you still revealing after the first 100 ticks?

ultraviolet-jordan:
> > Are you still revealing after the first 100 ticks?
> 
> https://github.com/2004Scape/Server/blob/832c4f25478387afd9ed1ba09bec388c194d9394/src/engine/World.ts#L1542
> 
> Here

`reveal` could probably be deleted or changed to something smaller like a `boolean` or even merged with another property?

mark-b5:
We set reveal to 100, and then decrement it every tick and compare it to 0. It acts as our reveal timer

Booleans are 64 bits anyway in typescript, no?

ultraviolet-jordan:
> We set reveal to 100, and then decrement it every tick and compare it to 0. It acts as our reveal timer
> 
> Booleans are 64 bits anyway in typescript, no?

Where is it decremented at?

ultraviolet-jordan:
> https://github.com/2004Scape/Server/blob/6af5422508773f6b39b5efeadd27cc96aed798d0/src/engine/entity/Obj.ts#L28-L29

Ah I see this just went in today I thought I was tweaking lmao sorry bout that.

PR #1640, opened 2025-04-08
refactor(content): gnome cooking

tannerdino:
- removes all of the varps used, and replaces them with 2 invs
- removes the very complicated code for the `inspect` op. We need to figure out what this does! For now I made it tell the player if the cocktail shaker or half-baked item is empty or not.
    The inspect op could be a number of things...
    - An ~mesbox listing the ingredients - this is a little annoying to code
    - A game message listing the ingredients - this is very unlikely I feel like (and was the previous code)
    - A ~p_choice with options of what ingredients to put next, listing the 'step' you're on?
    - A game message telling the player if the thing is empty or not (what I went with)
    - Or something entirely different.

I haven't found a lick of evidence as to what the `inspect` op should do :(


ultraviolet-jordan:
Why? Doesn’t it work just fine as is?

tannerdino:
It was way too complicated, I doubt it actually worked that way. The varps were used to store items like a inv, so the functionality is pretty much the same.

The inspect op, I'm still trying to think how it could've worked. IDK if appending to a string is something they would've done back then... But it could be an mesbox that shows each ingredient you put into it, in order

ultraviolet-jordan:
Are there any confirmed inventories that have a size of 7 and 11 respectively? The problem with inventing inventories to use is there will definitely be conflicts in the near future when the ones you picked are actually used for something else.

tannerdino:
No. Those came from 2009, which would've been after the gnome cooking rework in 2006. I assume they would've changed the inventories for some other use after the rework.

If they were changing the size of the inventory surely that means the invs were temp... What would even happen if you changed the size of a full inv

ultraviolet-jordan:
> No. Those came from 2009, which would've been after the gnome cooking rework in 2006. I assume they would've changed the inventories for some other use after the rework.
> 
> If they were changing the size of the inventory surely that means the invs were temp... What would even happen if you changed the size of a full inv

We really need to get our hands on an official 2004 server ong...... It's probably fine then if that's the case. It's too difficult to know.

ultraviolet-jordan:
And also what are you referring to when you say `inspect op` ?

tannerdino:
Screenshot_20250408-122723

ultraviolet-jordan:
> Screenshot_20250408-122723

yeah that's probably gonna be a gg's based on my experience watching every single video on youtube pre 2007. i will keep an eye out for this though if i ever find it.

tannerdino:
Something like this is relatively easy to code

https://github.com/user-attachments/assets/76ed4e0d-5df2-464a-9886-d0e1ade98982

tannerdino:
Since no invs were probably used, I went with ~p_choice, similar to post rework (just no special interface). We need some screenshots or videos of this :(

PR #1647, opened 2025-04-09
fix(engine): send tabs on reconnect

zeruth:
Tabs could have been updated while reconnecting, leaving clients in a broken tab state upon login

ie: disconnected while travelling to Entrana
image


ultraviolet-jordan:
Nice find, these older clients are a struggle.

PR #1663, opened 2025-04-14
fix(content): combat camp ogre check

Indio3:
Change behavior to match osrs, also fixes a bug with cannoning these


thesuddensilent:
How does that training_camp ogre thing work now as a whole after this fix?

Indio3:
> How does that training_camp ogre thing work now as a whole after this fix?

1-1 with osrs

PR #1666, opened 2025-04-14
chore(engine): Housekeeping for reset defaults

mark-b5:
Reverts a small change from https://github.com/2004Scape/Server/pull/1613 . Some content was relying on `npc_setmode(null)` to clean up a bunch of npc variables. The previous PR took that behavior away, but this PR adds it back

We might revisit this in the future and flesh it out a bit more. It's a little unintuitive that `npc_setmode(null)`  should have a bunch of side effects. It's possible that jagex uses a command like `npc_reset()` to achieve this behavior instead. Further, there's a few examples where we *do not* want this side effect, like failing a pickpocket, which is why the side effect was removed in the first place

For now, failing a pickpocket will reset a bunch of Npc state (inauthentically) until we can revisit this. This PR is essentially a temporary solution for tomorrow's update, because some content like gnomeball was relying on this behavior

I took this opportunity to rename `defaultMode` to `resetDefaults` also

### DIFF
--- a/src/engine/entity/Npc.ts
+++ b/src/engine/entity/Npc.ts
@@ -146,7 +146,7 @@ export default class Npc extends PathingEntity {
             }
 
             this.varsString.fill(undefined);
-            this.defaultMode();
+            this.resetDefaults();
 
             const npcType: NpcType = NpcType.get(this.type);
             this.huntrange = npcType.huntrange;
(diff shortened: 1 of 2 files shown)


mark-b5:
Also reset default huntrange when resetdefaults() is called @tannerdino

PR #1678, opened 2025-04-16
fix(content): Added `defaultmode=none` to more fishing spots

mark-b5:
Missed this file in the last PR, and it was the most important one 

Added defaultmode=none to all fishing spots, because npcs in wandermode will tele to their spawns


ultraviolet-jordan:
What exactly was it that caused them to start requiring these to be set on their config?

zeruth:
> What exactly was it that caused them to start requiring these to be set on their config?

NPCs resetting after 500 ticks, fishing spots should be excluded.

mark-b5:
^ 

We have better engine behavior for teleporting Npcs to their spawn tile after 500 ticks of failing to wander. We tested in osrs and this applies to Npcs in wander mode even if they have `moverestrict=nomove`, so we changed engine behavior to match that. This caused fishing spots to tele to their spawn constantly because they were all left in wander mode when their configs were made

ultraviolet-jordan:
tyty

PR #1679, opened 2025-04-16
fix(content): some levelup unlock messages

ultraviolet-jordan:
you can now try fly fishing for trout (2014)
you can now try fly fishing for salmon  members can catch leckish fish (2023)
----
You can see that Smithing had a change in these dagger messages where smelting was a newer addition:
you can now make steel daggers (2014)
you can now make mithril daggers (2006)
you can now make adamant daggers (2007)
you can now smelt iron ore and make iron daggers (2024)
you can now smelt steel and make steel daggers (2023)


ultraviolet-jordan:
I found some other ones to fix so I updated title and info.

PR #1680, opened 2025-04-17
fix(content): gnomeball adjusted to account for new engine behaviors

tannerdino:
Relies on https://github.com/2004Scape/Server/pull/1668 to be merged (resolved merge conflicts)

So with further testing I was able to refine our current gnomeball implementation. Our old implementation also relied on *old engine behavior*. A few things that didn't take into account new engine functionality!
- When a gnome winger passes the ball to the player, the *distance between the target and source* has a 3 tile variance. So the flight duration is seemingly "random". The theory goes: since the pass to the player p_delays the player for 1t, the coords are initialized before the p_delay. Meaning if the player or gnome winger is moving around, you'll get weird inconsistent flight durations.
- Gnome winger timer behavior was previously using a varn to keep track of when to throw the ball back. This was originally done because if you pass the ball back to the gnome winger right after he passes it to you, he'll always pass it within 10 ticks apart. I replicated this behavior by just setting the timer to 10 ticks when he passes back - not sure why I used a varn to do this. Gnome wingers also have a 1t timer on respawn, and the timer randomization is set *in the timer itself*
- The old code also had the gnome winger timer handle the "respawn" of gnome winger if you leave the game. Through new engine knowledge, this is simply just an `npc_changetype()` :D
- Some of our delays when tackling/passing were either too short or too long. Few other delays were mistimed as well. Confirmed through playtesting and event logging
- Gnome referee had some funky code to check if the gnomeball was still in play, that would conflict if there were multiple people playing. Instead I made it look for gnomes that have %npc_attacking_uid set to your uid. 
- Removes the foodshop inv, that the previous code was using for storing the gnomeball if you win (or entered with one). I use the gnome progress bits to keep track of this now.
- When in combat, the gnome tacklers will still tackle you! So I removed the check_notcombat from their hunt config
- Previous code had prevented gnome archers from attacking you when in a gnome game, which was just wrong when I retested in osrs
- In osrs theres a bug with gnomeball where if a gnome tackler has your ball, and you leave. Itll change their type back to their original type.... But it WONT turn their hunt back on. My previous code set %npc_lastcombat to some super high value to replicate this. When it likely just turned off their hunt when they steal your ball, and doesnt always turn it back on.

tldr: Authentic gnomeball behavior!


mark-b5:
Can probably clean up the git history a bit

PR #1687, opened 2025-04-18
fix(engine): Fixed ammo pickup bug and doors disappearing bug

mark-b5:
https://github.com/2004Scape/Server/pull/1683

There was a massive bug that was caused if a loc was changed and then `loc_changed` back to its original form. Added 2 failsafes for this


mark-b5:
Fixed a bug where deleting a dynamic loc would queue a pending event to delete the dynamic loc *again* erroneously in the future. The new debugproc `loc_del_event` will demonstrate the behavior if it is ran on main currently

PR #1689, opened 2025-04-20
fix(content): Ranged ammo poison fix

AmVoidGuy:
It was originally only checking the equipped weapon for poison instead of both like it does in npc ranged combat.


tannerdino:
This would affect poison knives, you'd want to check bi

tannerdino:
Missclick Im sorry, was reading this on mobile

PR #1693, opened 2025-04-21
fix(content): add authentic line breaks in random event dialogue

tannerdino:
image
image


mark-b5:
The old man text doesn't match the image. Is that on purpose?

mark-b5:
{EE8418A5-55A9-413E-BB74-3543748079FB}
this differs too

tannerdino:
Oh wow didn't even notice that typo! I'll add it in!

PR #1697, opened 2025-04-21
fix(content): remove members check from lava eel fishing to match osrs

tannerdino:
In osrs it just lets you fish in f2p, no members check at all

https://github.com/user-attachments/assets/18e874c2-1716-442d-879c-8a1445486b63


ultraviolet-jordan:
nice

thesuddensilent:
The Wilderness lava eel spot is not the same npc id as ours.      ..But I won't take a position on this...

ultraviolet-jordan:
That is expected bcuz osrs npcs "ids" got scrambled a few times...

PR #1700, opened 2025-04-22
fix(engine): INV_DROPALL crash due to misplaced brace

DrJacobHolden:
Bug:
Players who die while meeting the condition `wealthLog.size > 0` fail to die "correctly". Instead they are logged out. When they log back in they are in the same place as they died, they have 0 hitpoints and they have lost _all_ of their items.

Error message:
```
script error: inv_dropall wealthLog.values(...).map is not a function
file: death.rs2

stack backtrace:
    1: [proc,player_death_lose_items] - death.rs2:91
    2: [proc,player_death_lose_items] - death.rs2:63
```

Fix:
Re-arrange brackets to prevent the script from crashing.

### DIFF
--- a/src/engine/script/handlers/InvOps.ts
+++ b/src/engine/script/handlers/InvOps.ts
@@ -759,7 +759,7 @@ const InvOps: CommandHandlers = {
         }
 
         if (wealthLog.size > 0) {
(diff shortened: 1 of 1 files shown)


DrJacobHolden:
This is only an issue on unsupported Node versions (<22)

PR #1706, opened 2025-04-24
fix(engine): Item stacks will no longer reverse their order if a player leaves and re-enters a zone

mark-b5:
We were sending the order of Locs and Objs in the order of newest->oldest before this PR. This was an externality of the new zone impl. rolled out a month or two ago

I tested on rsprox and the order should be oldest->newest when reloading a zone

Bug report here: https://discord.com/channels/953326730632904844/1365080774705025095/1365080774705025095


mark-b5:
Pushed another update that rearranges Locs so that WriteFullFollows matches OSRS order

PR #1713, opened 2025-04-26
fix(content): earth, ice, and shadow warriors to no longer drop bones

tannerdino:
matches osrs, also runehq missing the 100% drop:
- [Earth warriors](https://web.archive.org/web/20040813232135/http://runehq.com/viewMonster.php?id=58)
- [Ice warriors](https://web.archive.org/web/20040813231551/http://runehq.com/viewMonster.php?id=50)


tannerdino:
image

They seemed to have dropped bones until mid 2004? (they also dropped bones in classic)

PR #1720, opened 2025-04-27
feat(content): 1/4 of NPCs named authentically

thesuddensilent:
First 252 npc names updated to source names.
(note that there are a few names here that are not found in the source list because the npc in question no longer exists. or it has been changed or more npc variations have been added so the name has changed.

--Guesses--
7=farmer
//osrs has several variations of this npc and the name is farmer1. the name was probably changed in 2005.
//the original name is probably just farmer.
8=thief
//osrs has several variations of this npc and the name is thief1. the name was probably changed in 2005.
//the original name is probably just thief.
10=guard2
//the npc was added in the falador graphics update in 2006 the npc in question was added.
//the name used on the osrs side is fai_falador_guard2. the original name is probably just guard2.
17=barbarian_woman
//this npc has been changed to a new npc but the equivalent is found in varrock from the bar but it's still not the same npc.
51=babydragon
65=leprechaun
69=lizardman
70=mummy
71=orc
72=troll
121=dwarf_mountain2
135=mammoth
170=gnomepilot
//this npc has been replaced in 2006 with unique npc ids.
--Guesses--

information taken from here.
https://chisel.weirdgloop.org/moid/npc_id.html
https://raw.githubusercontent.com/Joshua-F/osrs-dumps/refs/heads/master/config/dump.npc


AmVoidGuy:
It's a bit of a chore, but I think if you're going in and updating the names of them, it would be good practice to update the labels and file names as well so it's clear in the future in case something needs to be searched for.

For instance:
dwarf_youngster.rs2 -> dwarfchildtw1.rs2
@dwarf_youngster_dialogue -> @dwarfchildtw1_dialogue

thesuddensilent:
I understood this, I thought to myself that those should probably be changed too. But I just didn't dare, but if the package is bad, there's no need to add it.

PR #1723, opened 2025-04-29
Necromancer drop chance fix.

hugojmaia:
Necromancer only rolls a 20 instead of a 128, leaving most of the drop table unattainable.
This fixes it.


mark-b5:
https://github.com/2004Scape/Server/pull/1707 duplicate

PR #1743, opened 2025-05-07
chore(engine): Update map_members. npc_findhero and testbit to return int

Indio3:
image
image


tannerdino:
I think a bunch of other commands should return int as well.

from the list of runescript leaks it looks like:
- `testbit`
- `bufferfull`
- `oc_param`
- `inzone`
- `inarea`
- `huntnext`
- `lineofwalk`

strangely some use booleans:
- `loc_findnext`
- `npc_findexact`
- `region_findbycoord`
- `finduid`
- `controller_findexact`
- `loc_find`
- `huntnext` ??


They probably all return int... maybe boolean and int both work for them, or they were changed overtime

Indio3:
> I think a bunch of other commands should return int as well.
> 
> from the list of runescript leaks it looks like:
> 
> * `testbit`
> * `bufferfull`
> * `oc_param`
> * `inzone`
> * `inarea`
> * `huntnext`
> * `lineofwalk`
> 
> strangely some use booleans:
> 
> * `loc_findnext`
> * `npc_findexact`
> * `region_findbycoord`
> * `finduid`
> * `controller_findexact`
> * `loc_find`
> * `huntnext` ??
> 
> They probably all return int... maybe boolean and int both work for them, or they were changed overtime

i'll update the pr with the ones we can confirm return int (oc_param has a dynamic return, we also don't have bufferfull implemented yet)

PR #1751, opened 2025-05-09
fix(content): Loc Axis check on sheepherder gate

AmVoidGuy:


ultraviolet-jordan:
image

AmVoidGuy:
Update to check that you are only required to wear the plague outfit when you are entering the gate instead of both entering and leaving so you can't get trapped inside. This will make it match Osrs. 

Also the dialogue doesn't make sense for it to apply to both sides. So assumedly it has always been like that. 
```
"It doesn't look very safe in there. I'm not going in without decent protective clothing."
"This is a restricted area.|You cannot enter without protective clothing."
```

